PR #2, opened 2018-10-01
Implementation of the TrainedBayesianConsensusCaller

jeizenga:
I've filled in the boilerplate on the TrainedBayesianConsensusCaller. The extent of the testing I've performed is that I can get it to compile, so it's possible there are bugs in here. However, I gave Ryan a prototype of the same algorithm and it seemed to be working in my hands, so the algorithm is conceptually sound, at least. 

I also see that you've set up the API to return only a single consensus call. It would also be possible under this model to return a distribution over consensus calls, or at least a probabilistic confidence score. Let me know if you'd like me to add that functionality.


paoloczi:
Cool, thank you! When you have a chance, can you also add an input file with the current values of the training parameters? You can put it in the conf directory. At some point I will modify the documentation to note that that file must be copied or symbolic linked to the run directory.

Regarding the probability distribution: that is a very good thing, so please do it in the derived class. We will not be able to use it for now because there is no abstract interface for it, but let's start with that.

paoloczi:
This is great! It merged and built without problems. I will give it a try in a local assembly as soon as you add the input file with the parameters.

A couple of minor comments for consistency with the rest of the code - no code changes are needed:
* I use CZI_ASSERT instead of assert. It is always compiled in (not just in debug builds), and on failure it throws an exception, which ensures a cleaner termination and can be caught.
* For many common identifiers in the std namespace, I have files like string.hpp and vector.hpp which inject the name in the shasta namespace, so you don't need to qualify them with std::, which makes the code a bit more readable.

PR #3, opened 2018-11-16
linked to polisher library and added simple test executable

rlorigro:
edited cmakeLists to include external project MarginPolish, added test src, header, and executable for linkage "testLassenLink"

### DIFF
--- a/scripts/FlagMarkerGraphWeakEdges.py
+++ b/scripts/FlagMarkerGraphWeakEdges.py
@@ -12,7 +12,8 @@ a = shasta.Assembler()
 a.accessMarkerGraphVertices()
 a.accessMarkerGraphConnectivity(accessEdgesReadWrite=True)
 a.flagMarkerGraphWeakEdges(
-    minCoverage = int(config['MarkerGraph']['minEdgeCoverage']),
+    lowCoverageThreshold = int(config['MarkerGraph']['lowCoverageThreshold']),
+    highCoverageThreshold = int(config['MarkerGraph']['highCoverageThreshold']),
     maxDistance = int(config['MarkerGraph']['maxDistance']),
     )
 
--- a/src/Assembler.hpp
+++ b/src/Assembler.hpp
@@ -354,7 +354,8 @@ public:
 
     // Find weak edges in the marker graph.
     void flagMarkerGraphWeakEdges(
-        size_t minCoverage,
+        size_t lowCoverageThreshold,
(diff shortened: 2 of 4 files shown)


paoloczi:
Some comments: 

1. I don't want to add the test executable to the mainline Shasta code, as its presence could confuse users, in addition to adding an additional, unnecessary link step to all builds. Any chance that you could move that to a separate branch in your fork? That way you could still use it for your testing without forcing all Shasta builds to create it.
2. I thought we had agreed that, before we add the external project, you and/or Jordan would add #ifdef's in Benedict's code to compile only the portions that we need. If we don't do that, we add unnecessary dependencies to the Shasta code, and I don't want to do that. 
3. Could you add more comments to the portions of CMakeLists.txt that you added? For consistency with other comments, please capitalize the first letter of the first word and add a final period to each comment. It is ok/desirable if the comments are verbose and/or multi-line.

rlorigro:
1. Yes I can move these
2. Can also start on this, I assumed since it was a different repo, whatever work we do over there wouldn't influence this project's code. But I see what you mean about the dependencies. We could also add some sort of toggle flag that would opt-in for the polisher
3. Yes

rlorigro:
OK all 3 changes should be addressed now. I removed htslib and lastz from the polisher, as well as all user-installed libraries like bz2, curllib, etc.

Also simplified the linking process w.r.t. rpath. Turns out adding multiple rpaths is doable, so copying the external library into the install directory is no longer necessary

paoloczi:
Thank you for making these changes! 

I have a question on removing the dependencies of the marginPhase library. I don't see a special compilation definition when building that library. But I know the library does have those additional dependencies. How do you make sure we only compile the portions that are needed for our purpose, and that don't use the dependencies?

Also, I have a question about the way you link the additional library. I see that it is a static library. I never knew that it was possible to link a static library into a dynamic library, but obviously this is the case because you are doing it. This is neat, but in this case access to the static library is not needed at runtime - only at link time. Therefore the shasta library should only have $ORIGIN on its rpath, not the additional directory. 

If the marginPhase library were instead dynamic, you would still only want to have $ORIGIN and nothing else on the rpath, and instead copy the marginPhase library to the shasta install directory. The reason for this is that we want the shasta-install directory to remain self contained - that is, it will continue to work if it is moved to a different machine, or if the rest of the build directories are moved or removed.

We are getting close. As soon as you make the rpath change I will clone your fork and do some testing.

PR #4, opened 2018-12-08
Implementation of simple Bayesian consensus caller

rlorigro:
I have implemented the consensus caller class previously templated as SimpleBayesianConsensusCaller. A few things that are still pending:

- Test with actual read data (especially for gaps)
- Convert boolean flags to const type for flags that determine behavior regarding gaps (I have put off figuring this out for now). Right now they are hard coded during the constructor call.
- Use the same namespace conventions as the rest of shasta
- Find a proper place to store probability matrix csvs (current in shasta/models/)
- Convert to camelCase?
- Remove use of "this"?

Also, I have generated a normalized vector of log likelihoods for each possible "true" repeat, but at the moment it is unused in the operator function. As the template specified, only a "Consensus" object is returned with base and repeat count.


paoloczi:
This is great, and I don't think any of the tasks you list as pending should prevent us from merging this work. You or I can always fix those things in time.

But there are a few small things that I would like done before merging:

* Remove extraneous changes in CMakeLists.txt, src/CMakeLists.txt, src/PythonModule.cpp. I know these are residuals of the external project work.
* Change the code to read all the four matrices from a single csv file. At some points we should add header lines at the beginning of each matrix, but let's not worry about that for now. This should be easy to do - just open the ifstream in load_probability_matrices, and pass it as an argument by reference to load_probability_matrix (instead of the file_path), like this: 

```C++
load_probability_matrix(istream& matrix_file) {
```
* Modify load_probability matrix to pass a single file_path instead of a vector of paths. Ideally, this path should be added as a constructor argument instead of being stored as a constant.

* Rather than creating a new models directory, let's put that file in existing directory shasta/conf and call it SingleBayesianConsensusCaller-1.csv. It can be considered configuration information as it affects the assembler behavior. I will add code and documentation to allow using this file to use a SingleBayesianConsensusCaller for assembly instead of the current default SimpleConsensusCaller.

rlorigro:
- CMakeLists.txt, src/CMakeLists.txt, src/PythonModule.cpp
    - This should have been corrected already, but it might appear that it is not at first glance. I edited them twice: once for testing, and once to manually revert them. The commit labeled "removed test executable and reverted minor changes" should have handled all these, but now I have reverted each file via "git checkout" instead

- Change the code to read all the four matrices from a single csv file.
    - Done, using fasta-like headers: ">A", ">C", etc

- Modify load_probability matrix to pass a single file_path instead of a vector of paths. Ideally, this path should be added as a constructor argument instead of being stored as a constant.
    - I added a string argument "matrix_file_path"  which has a default value of the const path

- Rather than creating a new models directory, let's put that file in existing directory shasta/conf and call it SingleBayesianConsensusCaller-1.csv
    - Done

paoloczi:
This looks good. Just remove the four old csv files in the models directory and I will merge it.

rlorigro:
ok done, after some fiddling with git

paoloczi:
Thanks, cool! I will merge first thing tomorrow morning, after submitting some pending changes on my side.

PR #6, opened 2019-01-08
Cleaned version of SimpleBayesianConsensusCaller

rlorigro:
Ok here is the cleaned, error and warning free version of the consensus caller


rlorigro:
ah I see, should be fixed now. Maybe at some point that can be automated with some kind of user prompt

paoloczi:
Requesting user intervention is not good for things that could run in batch.

rlorigro:
Well, I mean during setup, not during runtime

PR #8, opened 2019-01-29
Added one shot script for small runs

rlorigro:
This is a simple argparser script that can run a small assembly using only the command line and the path to your Fasta file. All default conf params can be overridden by the user except useRunLengthReads. If not specified, values from the shasta.conf file in shasta-install/conf/ are used instead.

RunAssembly.py and SetupSmallRunDirectory.py have been edited so they are modular, and their functions can be called from neighboring scripts.


paoloczi:
Thanks - this looks good but I am going to do some testing myself. Once I check that everything works I will also add some documentation changes so users know that this additional option is available.

paoloczi:
Everything works great - I just had to make two small changes:

- Added a "#!" line at the beginning of SmallRun.py to prevent problems if it was invoked without prefixing with "python3".
- Put back a small change in RunAssembly.py I made yesterday and that was probably lost in a merge.

I also added some user documentation - you can review it
[here](https://chanzuckerberg.github.io/shasta/RunningSmall.html). 
If you want to propose changes to this, you can make changed in docs/RuniinigSmall.html and submit a pull request.

Just a minor suggestion: option name "--sequences" seems a bit generic. Perhaps use a different name that makes it clear that these are the input files containing reads? But I don't have a strong feeling on this.

rlorigro:
Awesome. Thanks for testing it.

- I think there's a mistake in the new documentation, as it says SetupSmallRunDirectory instead of SmallRun.

- I also don't have a strong feeling about the "sequences" flag. My only goal was not to name it after a file format, and not to suggest some kind of use case. inputSequences, reads, input, or something else would work too. "Reads" suggests it came from a sequencing machine, but the sequences could be synthetic, or preprocessed from some other consensus algorithm.

- at the moment I think it would fail if you tried to enter multiple input files. So I can update this if you want to retain that ability

paoloczi:
- Oh yes, I had noticed the error in the script name and I already made the change, but I did not submit it. It's in now.

- Perhaps "inputSequences"?

- It would be useful but not essential to make it work if there are multiple input files. I can certainly see scenarios for using multiple input files, but probably not for small runs. And it is relatively easy to combine input files anyway. Your choice on this.

rlorigro:
Ok I updated the flag names. I decided to just leave it as a single fasta file for now.

also switched all the `!= None` statements to `is not None`. I suppose we need to start a new pull request?

paoloczi:
Yes, please start a new pull request. Once merged, I don't know of an easy way to add additional changes.

PR #10, opened 2019-02-01
one shot script for full size runs 

rlorigro:
Hi Paolo, 

Here is what I have running on my machine. It is nearly the same script as SmallRun, but with added steps to allocate pages and later remove them or save them, based on your own preexisting scripts. If you don't mind testing it, that would be helpful (considering the possible heisenbug I observed earlier).


rlorigro:
ah sorry, I omitted the actual script from the commit by accident. Had too many temp files cluttering my git gui.

Should be there now as "Run.py". Also removed that comment.

paoloczi:
Thanks. Working on testing it now. A couple of other minor things that you could look at in the meantime:
* src/AssemblerAssemblyGraph.cpp is still in the list of changed files due to a extra blank line. Not essential, but it would be nice to remove that file from the pull request.
* I just noticed that Run.py and SmallRun.py are not executable like all the other scripts in the directory (use `ls -l shasta/script` to check). When you have a chance please make them executable (e. g. via `chmod ugo+x Run.py SmallRun.py`) so they can be invoked directly.

rlorigro:
ok I pushed another update. Hopefully the lines match now.

rlorigro:
I probably should have just used this method to revert it, instead of added more edits:

http://nicolasgallagher.com/git-checkout-specific-files-from-another-branch/

paoloczi:
No problem, this is fine. The top level script Run.py looks correct to me. I will let you know how the testing goes. I will test both on Ubuntu 16.04 and 18.04.

paoloczi:
Some relatively minor issues I found during testing:
* When `--inputSequences` specifies a path relative to the current directory (e. g. simply `Run.py --inputSequences abc.fasta`), the script does not work.
* When the script fails (for example as in the previous bullet), it does not clean up huge pages.

rlorigro:
yes, that's a good point about the cleanup. I will work on both of these

paoloczi:
Another minor thing: option `--savePageMemory` is very useful, but from the help message it is not clear what the value should be. I tried specifying `--savePageMemory True` and the script accepted that, but then there was a failure at the end of the run with the following message: `Exception: Data does not exist.`

Other than the relatively minor issues I reported the script worked without problems on Ubuntu 16.04. I will try 18.04 now.

paoloczi:
On Ubuntu 18.04 I am getting the failure `MarginPhase was not setup`. I will investigate.

rlorigro:
OK. The C++ binding and threading is also making it a challenge to handle interrupt signals. Trying to run CleanupHugePages at termination is giving some messy errors:

```
umount: /hugepages: target is busy.
rmdir: failed to remove '/hugepages': Device or resource busy
Cleaning up page memory... Done
terminate called after throwing an instance of 'std::runtime_error'
  what():  Error 2 opening MemoryMapped::Vector Data/AlignmentCandidates: No such file or directory
Aborted (core dumped)
```

paoloczi:
I don't think it is a signal handling issue. Your messages show that the Assembler object is still alive when you attempt to clean up the pages and umount the filesystem. The umount fails because the Assembler object has several files in the huge page filesystem open. 

To avoid this issue, you need to make sure the assembler object gets destroyed before you umount. Adding `del a` in Python code before calling the cleanup function should take care of this.

rlorigro:
ah that would be convenient. I was looking in the wrong places for how to release the resources. `del` makes sense.

paoloczi:
One more minor thing: if the user want to use marginPhase, you also need to copy the marginPhase config file to the run directory like you do for the Bayesian consensus caller. The file must be called MarginPhase.json.

But this is no the problem because we are not trying to turn it on.

rlorigro:
ok I think I've addressed all the issues not relating to exception handling

PR #19, opened 2019-05-07
Make accent color complement of primary color

csweaver:
Changed `accent` color from darker version of `primary` to complement of `primary`. This fixes the issuer where links and code looked too similar. It may come off a little bright, still so LMK if you would prefer a different color/style.

Screen Shot 2019-05-06 at 10 42 05 PM


paoloczi:
This works well - thank you!

PR #31, opened 2019-06-10
enable static build including spoa using cmake's externalproject_add

ekg:
When building shasta, I ran into some confusion about how to include spoa.

This makes it so that using the static build CMakelists.txt results in the automatic download and configuration of spoa.

### DIFF
--- a/src-static-executable/CMakeLists.txt
+++ b/src-static-executable/CMakeLists.txt
@@ -58,7 +58,18 @@ add_definitions(-DSHASTA_STATIC_EXECUTABLE)
 #    link_directories(/usr/local/Cellar/boost/1.69.0/lib)
 # endif(MACOS)
 
+include(${CMAKE_ROOT}/Modules/ExternalProject.cmake)
 
+ExternalProject_Add(spoa
+  GIT_REPOSITORY "https://github.com/rvaser/spoa.git"
+  GIT_TAG "0b9e56da3241d6a887d64adcfd1ebfae29810c04"
+  UPDATE_COMMAND ""
+  INSTALL_COMMAND "")
+ExternalProject_Get_property(spoa SOURCE_DIR)
+set(spoa_INCLUDE "${SOURCE_DIR}/include")
+ExternalProject_Get_property(spoa INSTALL_DIR)
+#set(spoa_INCLUDE "${INSTALL_DIR}/include")
+set(spoa_LIB "${INSTALL_DIR}/src/spoa-build/lib")
 
 # Specify the libraries to link with.
 if(MACOS)
@@ -69,12 +80,14 @@ else(MACOS)
     # For arcane reasons, statically linking with the pthread
     # library on Linux requires "--whole-archive". 
     target_link_libraries(
-        shasta-static-executable
-        atomic boost_system boost_program_options spoa
-        -Wl,--whole-archive -lpthread -Wl,--no-whole-archive)
+      shasta-static-executable
+      "${spoa_LIB}/libspoa.a"
+      atomic boost_system boost_program_options
+      -Wl,--whole-archive -lpthread -Wl,--no-whole-archive)
 endif(MACOS)
   
-  
+target_include_directories(shasta-static-executable PUBLIC
+  "${spoa_INCLUDE}")
     
 # Also request the glibc library to be linked statically. 
 # This is notsupported on macOS.   
@@ -85,4 +98,4 @@ endif(NOT MACOS)
 # The static executable goes to the bin directory.
 install(TARGETS shasta-static-executable DESTINATION shasta-install/bin)
 
-
+add_dependencies(shasta-static-executable spoa)


paoloczi:
Building the Shasta code has several dependencies in addition to the Spoa library. Shasta provides script `InstallPrerequisites-Ubuntu.sh` to install all of the dependencies, including Spoa. This is explained in the [documentation](https://chanzuckerberg.github.io/shasta/BuildingFromSource.html) under "How To", "Build the code from source".

Even if we add Spoa to the Cmake file as you suggest, users will still need to use `InstallPrerequisites-Ubuntu.sh` to install the remaining prerequisites, so the build process will not be any simpler. For this reason, I am not inclined to add this additional complexity to the Cmake file- I don't see the benefit of this additional complexity.

Also, I would prefer to keep the Cmake file simpler when possible, and instead add complexity to  `InstallPrerequisites-Ubuntu.sh` as necessary.

However, it is possible that I am not seeing some benefits of what your are proposing, so please feel free to respond with additional comments.

ekg:
This mechanism allows us to include specific commits from git repositories.
It's simpler when the repository has CMakeLists.txt, but will work with any
repo. The downloaded and built packages are put into your build directory,
not freely stored in the repository, and this can be made to update them
when you change the commit or tag that is referred to. From the user's
perspective, the build happens in one step. There isn't need to a prior
step to download and build the dependencies. It's simpler and cuts down on
documentation and maintenance.

I wasn't aware of the other build dependency download script. I guess
that's fine if you prefer. At least I think it makes sense to document this
in the README.

I was also unable to build when using the non static script. That's a
separate issue though.

paoloczi:
Thanks, I will leave it as is then. I prefer to keep the README very small, but it does contain a link to the full documentation, which includes a [section](https://chanzuckerberg.github.io/shasta/BuildingFromSource.html) on how to build from  source. Because of the availability of a static executable in the release, I anticipate that only a minority of users will want to build the code from source.

Please file a separate issue for your other build problem. Please verify that you are on an Ubuntu machine (16.04 or 18.04) and that you are following the directions in the documentation section referenced above.

ekg:
The script for installing prerequisites appears to require sudo. This isn't
possible in many contexts, including the one that I have been working in.

One advantage of setting things up inside of the cmake build trees is that
you can set up the dependencies without requiring root access.

If I setup the build to work this way would you consider the PR?

ekg:
I see that many of the installed items are essential for the full pipeline. To clarify, I am interested in getting the shasta build to work.

paoloczi:
The build _does_ work as currently implemented and documented, but it is true that running `InstallPrerequisites-Ubuntu.sh` requires root access. However, after this script runs once, builds can be done without requiring root access.

It would certainly be desirable to have a way to build without requiring root access. Note, however, that Shasta is best used on a dedicated machine, particularly for large assemblies: for example, optimizing assembly performance [requires root access](https://chanzuckerberg.github.io/shasta/Performance.html), and without that you can lose 30% or more in performance. In such a dedicated setting, the assumption of root access it not unreasonable.

If we could enable building the code without root access without adding additional complexity in the CMake file, I would be all for it. But I worry about the extra complexity in the CMake file. In particular, I worry that the extra complexity may make hard to add additional dependencies in the future, because we don't know the requirements of those future additional dependencies. And making the CMake file too complex could discourage forking, particularly by people who might not be very experienced with CMake and who may want to modify the build process to add their own additional dependencies. Currently, the CMake files are very simple, and any user with basic knowledge of CMake would be able to understand them and modify them.

Another way to say this is that CMake is for doing the build, and installing dependencies is a separate business that can add a lot of complexity and is perhaps best kept separate. Of course, we could agree to disagree on this.

So overall my inclination would be to leave things as they are, and keep in place the current requirement that, if you want to build the code, you need root access. Particularly because only a minority of the users need to rebuild - for most users just downloading the executable from a release would be sufficient.

The other possibility would be to create a new directory that does the build without requiring root access, but it seems to me that this has the potential to create unnecessary confusion.

paoloczi:
There were no comments in one week, so I am closing this. Feel free to reopen it or to create a new PR if more discussion is warranted or separate issues arise.

ekg:
I just ran into this problem again when attempting to build shasta.

Many of your users are not going to have root access. I think you should consider opening and merging this to remove the dependency management problem when building shasta from source. The externalproject pattern is sufficient to do so.

paoloczi:
Root access would be necessary anyway to install the required packages via `apt install`. So even with this proposed change somebody without root access would still not be able to build Shasta, unless all of the required packages are already installed. 

The way things are, root access is required only during installation of the prerequisites, not during build. A user without root access can request an administrator once to run `InstallPrerequisites-Ubuntu.sh`, and after that root access is not required during the build.

As I said in the discussioin above, I still prefer to use cmake for building only, and use a separate process to install all required dependencies, including spoa.

PR #63, opened 2019-10-10
Build for generating dynamic library with GPU acceleration.

yatisht:


dynamicExecutable/CMakeLists.txt
```diff
@@ -1,32 +1,54 @@
...
+    find_package(CUDA 10.0 REQUIRED)
+    set(CMAKE_CUDA_COMPILER_ENV_VAR "CUDACXX")
+    set(CMAKE_CUDA_HOST_COMPILER_ENV_VAR "CUDAHOSTCXX")
+    project(shastaDynamicExecutable LANGUAGES C CXX CUDA)
+endif(BUILD_FOR_GPU)
+
 # C++ dialect.
-add_definitions(-std=c++14)
+#add_definitions(-std=c++14)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
```
paoloczi: I find `add_definitions` more readable than `set(CMAKE_CXX_FLAGS`, and [cmake documentation](https://cmake.org/cmake/help/v3.0/command/add_definitions.html) says that `add_definitions` can be used to add any flag, not just `-D` definitions.  Is there a reason why you were forced to switch to `set(CMAKE_CXX_FLAGS`?

src/gpu/GPU.h
```diff
@@ -0,0 +1,10 @@
+#include "../timestamp.hpp"
+
+#define MAX_MARKERS_PER_READ (1 << 15)
+#define GPU_BATCH_SIZE (1 << 12)
+#define MIN_GPU_BATCH_SIZE (1 << 4)
```
paoloczi: Please use a SHASTA prefix fore the above macros, to prevent name collisions.

src/gpu/GPU.h
```diff
@@ -0,0 +1,10 @@
...
+
+void common_markers (uint64_t* read_pairs, uint32_t offset, int n);
```
paoloczi: Is it possible to move this to the shasta namespace? I want to keep additions to the global namespace to a minimum to reduce the potential for name collisions and also for clarity.

src/gpu/GPU.h
```diff
@@ -0,0 +1,10 @@
...
+extern "C" int initializeProcessors();
+extern "C" void alignBatchGPU (size_t n, size_t deviceId, size_t maxSkip, size_t num_pos, uint32_t* h_reads, uint64_t* h_read_pairs, uint32_t* h_alignments);
+extern "C" void shutdownProcessors(int nDevices);
```
paoloczi: Is `extern"C"` required for the above functions? If so, please add a `shasta_` prefix to the names. If not, please move to the `shasta` namespace.

paoloczi:
This looks great - thank you! Please take a look at the minor comments above.

dynamicExecutable/CMakeLists.txt
yatisht: Yes, since we're also using NVCC compiler for GPU code, add_definitions would set flags for both compilers and NVCC would throw an error for these.

src/gpu/GPU.h
yatisht: Sure.

src/gpu/GPU.h
yatisht: I am removing this function as I do not use it anymore.

src/gpu/GPU.h
yatisht: Yes, since these functions are compiled they are compiled with NVCC. I'll add the prefix.

dynamicExecutable/CMakeLists.txt
paoloczi: In that case, if possible it would be better to put all of the GPU code in a separate directory, and create a separate cmake file for that. It's not good that the regular cmake files have to worry about the GPU compiler.

yatisht:
Regarding dependencies, only CUDA installation is required. Assuming Ubuntu and x86_64, the instructions are provided here:
https://developer.nvidia.com/cuda-downloads?target_os=Linux&target_arch=x86_64&target_distro=Ubuntu&target_version=1604&target_type=runfilelocal

paoloczi:
Is the CUDA installation always required? Even when using the static executable?

yatisht:
It gets you the necessary drivers, so preferably. There are ways to get just the drivers installed separately, if not already installed, but I prefer this method.

dynamicExecutable/CMakeLists.txt
yatisht: That might be possible - I can try doing that.

paoloczi:
Ok regarding the CUDA installation. I will document this at the appropriate time.

dynamicExecutable/CMakeLists.txt
paoloczi: Please give it a try. I don't want to create a situation where cmake changes for the CPU code break the GPU build or vice versa.

dynamicExecutable/CMakeLists.txt
```diff
@@ -1,32 +1,54 @@
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.15)
```
paoloczi: I am also just noticing that you bumped up the required cmake version to 3.15. I assume this is required by the GPU compiler, but this change makes it impossible to build Shasta on Ubuntu 16.04, which has cmake 3.5.1 (I don't want to force users to install a different cmake version from what comes with their system). Even with `BUILD_FOR_GPU` turned off, the build would fail on Ubuntu 16.04 because of this. 

Creating a separate cmake file for the gpu compilations may fix this, if cmake lets us specify a different minimum version in the gpu subdirectory. Another way would be to not use new cmake features, but this might not be possible for the gpu build.

In summary: ok to say that when `BUILD_FOR_GPU` is on, Ubuntu 18.04 is required. But not when `BUILD_FOR_GPU` is off. It's too early drop build support for Ubuntu 16.04

PR #66, opened 2019-10-21
GPU bug fixes and optimizations

yatisht:
Also tested on C. elegans


src/AssemblerMarkers.cpp
```diff
@@ -87,6 +87,19 @@ size_t Assembler::getNumMarkers(ReadId readId, Strand strand)
 
     return numMarkers;
 }
+
```
paoloczi: I would prefer that we don't use bare pointers like that because it lends itself to memory bugs. I would prefer to create a `std::vector v`, and if you need the pointer use `v.data()`. In this case, the routine could also allocate the vector with the right size. Also, the vector destructor takes care of releasing the memory when it goes out of scope, decreasing the possibility of memory leaks. No need to change this for now, just keep it in mind for the future.

yatisht: Ok, I'll take care of it in the next commit.

PR #74, opened 2019-11-08
Accelerating disjoint set computation on GPU.

yatisht:


src/Assembler.hpp
```diff
@@ -285,6 +285,39 @@ class shasta::Assembler :
...
         size_t threadCount
     );
+    
+    // Loop over all alignments in the read graph
+    // to create vertices of the global marker graph.
+    // Throw away vertices with coverage (number of markers)
+    // less than minCoverage or more than maxCoverage.
+    // Also throw away "bad" vertices - that is, vertices
+    // with more than one marker on the same oriented read.
+    void createMarkerGraphVerticesGpu(
```
paoloczi: I would prefer to put this declaration under `#ifdef SHASTA_BUILD_FOR_GPU`. I know it does not hurt, but it is just for readability, to make it clear that this function is only compiled in in that case.

src/gpu/AssemblerAlign-gpu.cpp
```diff
@@ -124,8 +124,8 @@ void Assembler::computeAlignmentsGpu(
     cout << timestamp << "Creating alignment table." << endl;
     computeAlignmentTable();
 
-    cout << timestamp << "Shutting down processors." << endl;
-    shasta_shutdownProcessors(nDevices);
+    //cout << timestamp << "Shutting down processors." << endl;
```
paoloczi: Fine to leave the message commented out, but nevertheless I would make it less ambiguous by changing "processors" to "GPUs" or something else. Most people would normally interpret "processors" as meaning "CPUs".

src/gpu/AssemblerAlign-gpu.cpp
```diff
@@ -124,8 +124,8 @@ void Assembler::computeAlignmentsGpu(
...
+    //shasta_shutdownProcessors(nDevices);
```
paoloczi: Why are we not shutting down the processors?

paoloczi:
I made some small comments that you can address at your convenience, but  I will merge the PR anyway.

src/gpu/AssemblerAlign-gpu.cpp
yatisht: They are now shut down later on (after disjoint set computation).

PR #85, opened 2020-01-08
configurable install target

ekg:
These updates allow shasta to be installed into a typical directory structure rooted in `DESTDIR`.

For instance:

```
# build shasta
cmake -H. -Bbuild && cmake --build build -- -j 4
# install shasta
make DESTDIR=boom install
cd boom
find
```

This yields the following structure:

```
./usr
./usr/local
./usr/local/share
./usr/local/share/shasta
./usr/local/share/shasta/conf
...
./usr/local/share/shasta/docs
...
/usr/local/share/shasta/docker
...
./usr/local/share/shasta/scripts
...
./usr/local/lib
./usr/local/lib/shasta.so
./usr/local/lib/shasta.a
./usr/local/bin
./usr/local/bin/shasta
./usr/local/bin/shastaDynamic
```

This update will make packaging shasta for distributions easier.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,9 @@ docs/doxygen/latex
 .idea
 
 cmake-build-debug/
+
+\#*
+.#*
+*~
+build/
+.gdb_history
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,20 +125,39 @@ if(BUILD_DYNAMIC_EXECUTABLE)
     add_subdirectory(dynamicExecutable)
 endif(BUILD_DYNAMIC_EXECUTABLE)
 
-
+if(NOT SYSTEM_INSTALL)
+  # Install to the shasta-install directory.
+  set(CMAKE_INSTALL_PREFIX .)
+endif()
 
 # Install the scripts.
 file(GLOB SCRIPTS scripts/*.py scripts/*.sh)
-install(PROGRAMS ${SCRIPTS} DESTINATION share/shasta/scripts)
+if(NOT SYSTEM_INSTALL)
+  install(PROGRAMS ${SCRIPTS} DESTINATION shasta-install/bin)
+else()
+  install(PROGRAMS ${SCRIPTS} DESTINATION share/shasta/scripts)
+endif()
 
 # Install the configuration files.
-install(DIRECTORY conf DESTINATION share/shasta USE_SOURCE_PERMISSIONS)
+if(NOT SYSTEM_INSTALL)
+  install(DIRECTORY conf DESTINATION shasta-install USE_SOURCE_PERMISSIONS)
+else()
+  install(DIRECTORY conf DESTINATION share/shasta USE_SOURCE_PERMISSIONS)
+endif()
 
 # Install the docs directory.
-install(DIRECTORY docs DESTINATION share/shasta)
+if(NOT SYSTEM_INSTALL)
+  install(DIRECTORY docs DESTINATION shasta-install)
+else()
+  install(DIRECTORY docs DESTINATION share/shasta)
+endif()
 
 # Install the docker directory.
-install(DIRECTORY docker DESTINATION share/shasta)
+if(NOT SYSTEM_INSTALL)
+  install(DIRECTORY docker DESTINATION shasta-install)
+else()
+  install(DIRECTORY docker DESTINATION share/shasta)
+endif()
 
 # The targets built in each subdirectory are
 # installed by the cmake file of each subdirectory.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,21 +127,18 @@ endif(BUILD_DYNAMIC_EXECUTABLE)
 
 
 
-# Install to the shasta-install directory.
-set(CMAKE_INSTALL_PREFIX .)
-
 # Install the scripts.
 file(GLOB SCRIPTS scripts/*.py scripts/*.sh)
-install(PROGRAMS ${SCRIPTS} DESTINATION shasta-install/bin)
+install(PROGRAMS ${SCRIPTS} DESTINATION share/shasta/scripts)
 
 # Install the configuration files.
-install(DIRECTORY conf DESTINATION shasta-install USE_SOURCE_PERMISSIONS)
+install(DIRECTORY conf DESTINATION share/shasta USE_SOURCE_PERMISSIONS)
 
 # Install the docs directory.
-install(DIRECTORY docs DESTINATION shasta-install)
+install(DIRECTORY docs DESTINATION share/shasta)
 
 # Install the docker directory.
-install(DIRECTORY docker DESTINATION shasta-install)
+install(DIRECTORY docker DESTINATION share/shasta)
 
 # The targets built in each subdirectory are
 # installed by the cmake file of each subdirectory.
--- a/dynamicExecutable/CMakeLists.txt
+++ b/dynamicExecutable/CMakeLists.txt
@@ -59,7 +59,11 @@ set_target_properties(shastaDynamicExecutable PROPERTIES OUTPUT_NAME "shastaDyna
 
 # This is needed to make sure the executable looks for the shared
 # library in the same directory.
-set_target_properties(shastaDynamicExecutable PROPERTIES INSTALL_RPATH "\$ORIGIN")
+if(NOT SYSTEM_INSTALL)
+  set_target_properties(shastaDynamicExecutable PROPERTIES INSTALL_RPATH "\$ORIGIN")
+else()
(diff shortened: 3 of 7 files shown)


paoloczi:
For simplicity and to avoid confusing users I prefer to leave it the way it is. I would be happy to merge equivalent functionality implemented as a script that the user can optionally run after `make install` and which moves the various pieces to where you want them. Once that is available I can update the documentation on building Shasta to mention that script and its functionality.

ekg:
Would it be useful to be able to install shasta in any root that the user wants? I think so. That's what I'm working towards here, because it's useful to me and will certainly be useful to package maintainers who pull this into debian or another system.

The non-standard installation directory and hard-coded RPATH in the dynamic executable are going to cause problems for anyone trying to package this software. My point is that eventually, someone else is going to make this patch or something like it.

The goal here is that the user identifies the output directory root and that's where things go. That could continue to be shasta-install. I think it'd be better to have a script which moves the files from their normal places into shasta-install, and let cmake install things into their typical places by default. IMO the current situation is more confusing to users. I've never even used the shasta-install directory.

The backstory here is that I'm packing shasta for guix. This will provide reproducible builds that can be referred to in publications. It will also make it easy for me to deploy on various non-Ubuntu 18.04 systems, build docker images including it, or run it on any HPC system where the system administrators are running guix. This is the package definition: https://github.com/ekg/guix-genomics/blob/master/shasta.scm.

ekg:
To be clear, this doesn't resolve the RPATH issue. That's something I've fixed in the packaging. There might be a resolution for it that plays nice with cmake's install method though.

paoloczi:
I can see that the different directory structure would be useful in some cases, and like I said I do not object to creating a script that moves things around as desired. If you implement such a script (in `shasta/scripts`), I will be happy to merge it and document it. However, I prefer not to make this change in the `cmake` files. From your point of view, using the script just means that you have to invoke the script after `make install`. Moving a few MB around would be essentially instantaneous.

The `RPATH` of the dynamic executable is set to `$ORIGIN`, which is a special value meaning that the loader will look for libraries in the same directory containing the executable. This is done so the loader can locate `shasta.so`. Everything will continue to work after things are moved, as long as the dynamic executable and the dynamic library continue to be in the same directory. All other libraries in standard locations on the loader search path will continue to be found. So there is no issue with `RPATH` as long as the executable and the libraries stay together.

With your proposed directory structure, you separate the library from the executable. Even in that case, you can use standard Linux commands to modify the `RPATH` of the executable, if necessary. Although most likely this would not be necessary because your `/usr/local/lib` directory would probably be in the library search path anyway.

Regarding reproducible builds that can be referred to in publications: I strongly advise against that as there are already Shasta releases, each with its own `git` tag, which fill this role. Having two possible sources of Shasta versions would create confusion and serve no useful purpose.

And regarding Docker functionality, Shasta already provides minimal Docker functionality. If you intend to develop that, it would be best to coordinate with the UCSC group that developed that functionality, to avoid duplication and user confusion.

ekg:
Modifying the cmake definitions simplifies packaging and reduces possible
errors that might occur. It also eliminates the need to synchronize this
second build install script.

What if I add a variable flag to enable the kind of install pattern I have
here? If that isn't specified, then the install will work exactly as
before. This will modify the cmake scripts but it won't have any effect on
your standard build process. It won't take any effort to maintain.

ekg:
Also, as for bit-for-bit reproducible builds, currently we need to identify
which operating system version and other dependency versions was used in
order to achieve the shasta build. Practically, this probably isn't a major
issue, but it is something that can be solved through appropriate packaging
or build systems.

For me, it's important to be able to install shasta in systems where users
don't have root. Guix is an emerging solution to this that makes it very
comfortable for users to install packages and work in the full system. This
avoids the problem of needing to drop into docker containers, which have
added system administration overheads (like mounting volumes, maintaining
filesystem state, etc.).

In some systems, like singularity and kubernetes, the whole set of
executables needs to be available in a single container. In this case, the
user needs to build that container, which can often introduce a kind of
dependency hell in which different required components cannot be combined.
For instance, one system may depend on Ubuntu 16.04, while another 18.04,
and neither can be built for the other. This makes the generation of a
single container very tricky. While the authors of these packages fulfilled
the need to make a future proof distribution of their software by making a
docker image, that image lives alone. Systems like guix (and I assume nix,
although I am not as familiar with it) provide a mechanism to compose
different packages with distinct dependency trees that might conflict into
a single system image. In effect each package is a kind of appimage that
contains the set of files that would be installed on the system.

(This seems kind of fantastic but is achieved by standard Unix methods. The
trick is that there is a fixed UUID like full path for each package root
that starts with /gnu/store/... and contains a hash of the package
contents. Things like RPATH can then be set relative to this stable path.
Everything needed for a particular environment can be composed with
symlinks.)

This is definitely a tangent but also maybe relevant to the UCSC docker
folks.

ekg:
The commits I have pushed make it so that the default behavior of the build does not change. This will leave everything as normal in your build and testing process.

The systemwide installation pattern is achieved by setting a cmake variable, like this:

```
cmake -DSYSTEM_INSTALL=1 -H. -Bbuild && cmake --build build -- -j 4
```

I have also resolved the RPATH setting issue that I mentioned so that shastaDynamic will look in ../lib for the shasta.so shared library.

Provided that this is noted in the documentation, I assume that this will save work for other people packaging the build.

druvus:
I would also like to see those changes to simplify packaging.  I would like to add it to Bioconda for easy installation.

paoloczi:
Like I said, I have no objection to adding a script that creates the directory structure you want. All that you desire can be achieved in this way. If somebody wants to create a Shasta distribution, it is not to much to ask to do `make install; createNewDirectoryStructure.sh` (but feel free to pick the name of the script).

I only want to add to the `cmake` files capabilities that cannot be achieved in any other way. For example, I provide a separate installation script that installs all of the dependencies. My rationale for that is that `cmake` files have a tendency to become unreadable, unmanageable, and hard to modify as software matures. The Shasta `cmake` files already have several necessary flags and several build directories. So, if additional complexity is desired, I will only add it there if there is no alternative.

In this case, a post-install script is a simple alternative that provides all of the desired functionality without adding complexity to the `cmake` files. Like I said, I would be happy to merge and document such a script.

ekg:
> I only want to add to the cmake files capabilities that cannot be achieved in any other way.

The specific capability that this PR gives you is correctly lining up all your install targets, setting your dynamic executable RPATH, and so on for a systemwide installation without any non-standard tricks or additional scripts. That's something you get for free with cmake if you follow some basic conventions. It requires no additional lines of code unless you want variable behavior.

> For example, I provide a separate installation script that installs all of the dependencies. 

This works but is apparently limited to Ubuntu and users with root. It installs a lot of things systemwide that are only needed for this build. It is helpful in describing the dependency set well enough to reproduce the build, but it's not really a usable solution in most situations that I work in.

> My rationale for that is that cmake files have a tendency to become unreadable, unmanageable, and hard to modify as software matures. The Shasta cmake files already have several necessary flags and several build directories. So, if additional complexity is desired, I will only add it there if there is no alternative.

I'm specifically trying to reduce the complexity of your build's install process to make it more maintainable. Adding an additional script to fix up the complexity would seem to go directly against your objectives, because it would add an additional file to maintain and synchronize with the build. Why define something two places when it can be defined in one?

The fact that @druvus has already chimed in suggests that dealing with build irregularity isn't an uncommon problem.

paoloczi:
I am closing this due to lack of additional discussion. If you still want to pursue this and new elements emerge, feel free to reopen it.

ekg:
@paoloczi all good to close if you don't want to merge this.

I think the remaining problem (if you use a separate script for installation) is setting up the dynamic executable RPATH. This will have to be configured in the build, implying modification of the cmake config.

It's important to note that (in the first version of this path) can set up the install target root with the `CMAKE_INSTALL_PREFIX` variable in addition to `DESTDIR`.

Maintaining an external patchset for shasta is acceptable if you're not willing to merge in changes for packaging maintenance. Keep in mind that this will be replicated by all the people who do packaging, for each system, while merging something similar to the changes which I was suggesting here and documenting the setup would minimize that work.

paoloczi:
As I explained before, there is no issue with RPATH. The existing RPATH value, `$ORIGIN`, allows the loader to locate `shasta.so` in the same directory where the executable is. As long as the library and the executable are moved together, all is good and the RPATH does not need to be modified.

With your proposed directory structure, the library moves to a separate directory which would be one of the standard search directories for the loader, so with that structure too the library would be found, without any need to change the RPATH. In that case the RPATH would not be needed, but also would not hurt.

In any event, the packaging script could easily modify or remove the RPATH to suit any needs using standard Linux command `patchelf`.

PR #88, opened 2020-01-09
Added more conditionals to catch empty strings/vectors

rlorigro:
This should account for the remaining cases where an out-of-bounds address may be accessed

### DIFF
--- a/src/SimpleBayesianConsensusCaller.cpp
+++ b/src/SimpleBayesianConsensusCaller.cpp
@@ -60,7 +60,6 @@ void validate_text_file(string configPath){
             throw runtime_error("ERROR: unprintable character detected in bayesian config file"
                                 "\n in config file: " + configPath +
                                 "\n at position " + to_string(c));
-
         }
         c++;
     }
@@ -277,6 +276,11 @@ void SimpleBayesianConsensusCaller::loadConfiguration(ifstream& matrixFile){
     string separators = " ";
 
     while (getline(matrixFile, line)){
+        // Ignore empty lines
+        if (line.empty()){
+            continue;
+        }
+
         // Header line (labeled via fasta-like headers)
         if (line[0] == '>'){
             vector<string> tokens;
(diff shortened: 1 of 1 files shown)


src/SimpleBayesianConsensusCaller.cpp
```diff
@@ -288,10 +292,10 @@ void SimpleBayesianConsensusCaller::loadConfiguration(ifstream& matrixFile){
             if (tokens[0] == "Name"){
```
paoloczi: Do we now here that `tokens` is not empty?

src/SimpleBayesianConsensusCaller.cpp
```diff
@@ -288,10 +292,10 @@ void SimpleBayesianConsensusCaller::loadConfiguration(ifstream& matrixFile){
...
                 parseName(matrixFile, line);
 
-            }else if (tokens[1] == "prior"){
+            }else if (tokens.size() > 1 and not tokens[0].empty() and tokens[1] == "prior"){
                 parsePrior(matrixFile, line, tokens);
 
-            }else if (tokens[1] == "likelihood"){
+            }else if (tokens.size() > 1 and not tokens[0].empty() and tokens[1] == "likelihood"){
                 parseLikelihood(matrixFile, line, tokens);
             }
```
paoloczi: Should we also add a check for an extraneous, not recognized line?

src/SimpleBayesianConsensusCaller.cpp
rlorigro: I believe the tokenizer will always return something if given a non-empty string, which is guaranteed in this case. If it contains no separators, it will just return the string as one token.

src/SimpleBayesianConsensusCaller.cpp
rlorigro: I thought about this, and I also considered adding an initial check to make sure that each expected header is found only once, using a set. It's up to you, I think. Maybe a conventional config parser library would have been best

paoloczi: Yes, at some point perhaps a overhaul of this parsing code will be called for. Let's not worry about this for now. 

I will merge this PR in a few days, after the upcoming release.

PR #101, opened 2020-03-23
Unbreak mac builds

bagashe:
Used explicit type casting to make LLVM happy. 

Test Plan: Verified that Github actions workflow was able to build on all supported platforms.


bagashe:
Yikes! Botched it again. Sorry.

PR #106, opened 2020-03-31
Implementation of a compression scheme for storing alignments.

bagashe:
Test Plan: Created Python bindings for a test wrapper method and tested in Python3 shell.

### DIFF
--- a/src/compressAlignment.cpp
+++ b/src/compressAlignment.cpp
@@ -23,32 +23,45 @@ void shasta::compress(const Alignment& alignment, string& s)
         const int32_t skip0 = int32_t(ordinals[0]) - int32_t(ordinal0);
         const int32_t skip1 = int32_t(ordinals[1]) - int32_t(ordinal1);
 
+        ordinal0 = ordinals[0];
+        ordinal1 = ordinals[1];
+
         // Find the end of the streak.
-        uint32_t n = 0;
+        uint32_t n = 1;
         for(uint64_t j=i+1; j<alignment.ordinals.size(); j++, n++) {
+            const auto& nextOrdinals = alignment.ordinals[j];
+            if(nextOrdinals[0] != ordinal0 + 1) {
+                break;
+            }
+            if(nextOrdinals[1] != ordinal1 + 1) {
+                break;
+            }
             ++ordinal0;
             ++ordinal1;
-            const auto& nextOrdinals = alignment.ordinals[j];
-            if(nextOrdinals[0] != ordinal0) {
-                break;
-            }
-            if(nextOrdinals[1] != ordinal1) {
-                break;
-            }
         }
 
+        i += n;
 
+        // cout << "(Skip0, Skip1, N) = (" << skip0 << ", " << skip1 << ", " << n << ")" << endl;
         // Serialize skipx, skipy, n into the output string
         // using the most compact format possible.
         if(Format0::ok(skip0, skip1, n)) {
             Format0 format(skip0, skip1, n);
             s.append(reinterpret_cast<char*>(&format), sizeof(format));
-
         } else if(Format1::ok(skip0, skip1, n)) {
             Format1 format(skip0, skip1, n);
             s.append(reinterpret_cast<char*>(&format), sizeof(format));
+        } else if(Format2::ok(skip0, skip1, n)) {
+            Format2 format(skip0, skip1, n);
+            s.append(reinterpret_cast<char*>(&format), sizeof(format));
+        } else if(Format3::ok(skip0, skip1, n)) {
+            Format3 format(skip0, skip1, n);
+            s.append(reinterpret_cast<char*>(&format), sizeof(format));
+        } else if(Format4::ok(skip0, skip1, n)) {
+            Format4 format(skip0, skip1, n);
+            s.append(reinterpret_cast<char*>(&format), sizeof(format));
         } else {
-            throw runtime_error("Unable to compress alignment streak: "
+            throw std::runtime_error("Unable to compress alignment streak: "
                 "skip0 " + to_string(skip0) +
                 ", skip1 " + to_string(skip1) +
                 ", n " + to_string(n));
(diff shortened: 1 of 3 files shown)


src/compressAlignment.hpp
```diff
@@ -77,26 +77,32 @@ that can be represented                    3       7      511   -2^19   -2^31
...
+
+        uint8_t extractFormatIdentifier(const char);
     }
 }
 
 
 
 class shasta::compressAlignment::Format0 {
 public:
+    static const uint8_t ID = 0b0;
```
paoloczi: I don't use uppercase for static constants (`ID, ID_MASK`). I use the same style as for other constants (`id, idMask`).

paoloczi: Any special reason to use the unusual bit format for this constant? How about just `0`?

src/compressAlignment.hpp
```diff
@@ -192,22 +210,32 @@ class shasta::compressAlignment::Format2 {
...
-         SHASTA_ASSERT(skip0Argument <= 511);
-         SHASTA_ASSERT(skip1Argument >= -512);
-         SHASTA_ASSERT(skip1Argument <= 511);
-         SHASTA_ASSERT(nArgument <= 512);
+        int32_t skip0Argument,
+        int32_t skip1Argument,
+        uint32_t nArgument):
+        formatIdentifier(ID)
+    {
+        SHASTA_ASSERT(skip0Argument >= -512 && skip0Argument <= 511);
```
paoloczi: Or use a call to `ok`?

src/compressAlignment.cpp
```diff
@@ -23,32 +23,45 @@ void shasta::compress(const Alignment& alignment, string& s)
...
+            s.append(reinterpret_cast<char*>(&format), sizeof(format));
+        } else if(Format3::ok(skip0, skip1, n)) {
+            Format3 format(skip0, skip1, n);
+            s.append(reinterpret_cast<char*>(&format), sizeof(format));
+        } else if(Format4::ok(skip0, skip1, n)) {
+            Format4 format(skip0, skip1, n);
+            s.append(reinterpret_cast<char*>(&format), sizeof(format));
         } else {
-            throw runtime_error("Unable to compress alignment streak: "
+            throw std::runtime_error("Unable to compress alignment streak: "
```
paoloczi: This message would not be understandable by a user. Ok to leave as is, but add something to the effect "this is a Shasta internal error...". Or else use `SHASTA_ASSERT`.

src/compressAlignment.cpp
```diff
@@ -57,10 +70,145 @@ void shasta::compress(const Alignment& alignment, string& s)
...
+{
+    uint32_t ordinal0 = 0;
+    uint32_t ordinal1 = 0;
+    
+    size_t pos = 0;
+    while (pos < s.size()) {
+        uint32_t streakSize;
+        int32_t skip0, skip1;
+
+        uint8_t fid = shasta::compressAlignment::extractFormatIdentifier(s[pos]);
```
paoloczi: I prefer long names (`formatIdentifier` instead of `fid`) but this is subjective and not a requirement.

paoloczi: No need to prefix with `shasta::` since we are in the same namespace.

src/compressAlignment.cpp
```diff
@@ -57,10 +70,145 @@ void shasta::compress(const Alignment& alignment, string& s)
...
+    Alignment decompressedAlignment;
+    shasta::decompress(compressed, decompressedAlignment);
+
+    // cout << "---Decompressed Alignment -----" << endl;
+    // for(const auto& x: decompressedAlignment.ordinals) {
+    //     cout << x[0] << ", " << x[1] << endl;
+    // }
+    // cout << "-------------------------------" << endl;
+
+    SHASTA_ASSERT(alignment.ordinals.size() == decompressedAlignment.ordinals.size());
```
paoloczi: These four lines can be written more compactly as 

`SHASTA_ASSERT(alignment.ordinals == decompressedAlignment.ordinals);`

src/compressAlignment.hpp
```diff
@@ -77,26 +77,32 @@ that can be represented                    3       7      511   -2^19   -2^31
 
 // Standard library.
 #include "string.hpp"
-#include "span.hpp"
 
 
 namespace shasta {
     void compress(const Alignment&, string&);
-    void decompress(span<const char>, Alignment);
+    void decompress(const string&, Alignment&);
```
paoloczi: We also need the signature with `span` because the bytes will come from a `MemoryMapped::VectorOfVectors`. That can be the lower level routine. The overload that takes a `const string& s` can call that passing `span<const char>(s.begin(), s.begin() + s.size())`.

src/compressAlignment.cpp
bagashe: Good point. Will change to `SHASTA_ASSERT`.

src/compressAlignment.cpp
bagashe: Good point. Will change.

src/compressAlignment.hpp
bagashe: Yup. Will add it.

src/compressAlignment.hpp
bagashe: Cool.

bagashe: It increases readability of what the expected format identifying bits are. Of course, that's a personal preference. I don't feel strongly about it.

src/compressAlignment.hpp
bagashe: You mean `SHASTA_ASSERT(ok(skip0Arg, skip1Arg, nArgument));`?

paoloczi: Yes.

src/compressAlignment.hpp
paoloczi: Either way is fine with me. Just curious.

PR #108, opened 2020-04-06
Compress and store good alignments.

bagashe:
On tested data sets, a compression of about 11.23X was observed.

### DIFF
--- a/src/AssemblerAlign.cpp
+++ b/src/AssemblerAlign.cpp
@@ -240,6 +240,9 @@ void Assembler::computeAlignments(
     // If true, discard containment alignments.
     bool suppressContainments,
 
+    // If true, store good alignments in a compressed format.
+    bool storeAlignments,
+
     // Number of threads. If zero, a number of threads equal to
     // the number of virtual processors is used.
     size_t threadCount
@@ -270,6 +273,7 @@ void Assembler::computeAlignments(
     data.downsamplingFactor = downsamplingFactor;
     data.bandExtend = bandExtend;
     data.suppressContainments = suppressContainments;
+    data.storeAlignments = storeAlignments;
 
     // Adjust the numbers of threads, if necessary.
     if(threadCount == 0) {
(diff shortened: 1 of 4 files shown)


src/AssemblerAlign.cpp
```diff
@@ -311,9 +311,13 @@ void Assembler::computeAlignments(
         }
 
         if (data.storeAlignments) {
-            const vector<string>& threadCompressedAlignments = data.threadCompressedAlignments[threadId];
-            for(const string& compressed: threadCompressedAlignments) {
-                compressedAlignments.appendVector(compressed.c_str(), compressed.c_str() + compressed.size());
+            const auto threadCompressedAlignments = data.threadCompressedAlignments[threadId];
```
paoloczi, line 314: This makes a copy. Use `const auto&` instead.

src/AssemblerAlign.cpp
```diff
@@ -311,9 +311,13 @@ void Assembler::computeAlignments(
...
+            const auto size = threadCompressedAlignments->size();
+            for(size_t i=0; i<size; i++) {
+                compressedAlignments.appendList(
```
paoloczi, line 317: Use `appendVector` instead for better performance.

src/AssemblerAlign.cpp
```diff
@@ -447,7 +458,10 @@ void Assembler::computeAlignmentsThreadFunction(size_t threadId)
             if (storeAlignments) {
                 string compressed;
                 shasta::compress(alignment, compressed);
-                threadCompressedAlignments.push_back(compressed);
+                thisThreadCompressedAlignments.appendList(
```
paoloczi, line 461: Use `appendVector` for better performance.

src/AssemblerAlign.cpp
paoloczi: Actually, this is just a pointer, so not a problem.

PR #109, opened 2020-04-07
Change title of marker frequency table on reads page to indicate whether a selected portion of the read is being visualized.

bagashe:


bagashe:
image

PR #110, opened 2020-04-09
Use stored alignments if available

bagashe:
Here's the speed up (time saved) for different options. 15.5% time saved across the board and about 18-20% improvement in compute speed.

714s -> 600s. (disk/ssd, filesystem)
633s -> 533s (memory 4k pages, anonymous)
629s -> 533s (memory 2m pages, anonymous)
629s ->  524s (memory 2m pages, filesystem)

333s -> 282s (memory 2m pages, filesystem, alignMethodFor*Graph 3)

### DIFF
--- a/src/AssemblerAlign.cpp
+++ b/src/AssemblerAlign.cpp
@@ -333,9 +333,9 @@ void Assembler::computeAlignments(
 
     cout << timestamp;
     if (data.storeAlignments) {
-        cout << "Stored compressed alignments." << endl;
+        cout << "Cached compressed alignments for potential reuse." << endl;
     } else {
-        cout << "Not storing alignments." << endl;
+        cout << "Not caching compressed alignments." << endl;
     }
 }
 
--- a/src/AssemblerMarkerGraph.cpp
+++ b/src/AssemblerMarkerGraph.cpp
@@ -2,6 +2,7 @@
 #include "Assembler.hpp"
 #include "AlignmentGraph.hpp"
 #include "ConsensusCaller.hpp"
+#include "compressAlignment.hpp"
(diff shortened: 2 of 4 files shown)


src/AssemblerAlign.cpp
```diff
@@ -333,9 +333,9 @@ void Assembler::computeAlignments(
 
     cout << timestamp;
     if (data.storeAlignments) {
-        cout << "Stored compressed alignments." << endl;
+        cout << "Cached compressed alignments for potential reuse." << endl;
```
paoloczi, line 336: I prefer "storing" instead of "caching" here. "Caching" gives the impression of something temporary and dynamic, which is not the case here.

src/AssemblerMarkerGraph.cpp
```diff
@@ -489,43 +494,46 @@ void Assembler::createMarkerGraphVerticesThreadFunction1(size_t threadId)
...
-                    alignment, alignmentInfo
-                );
-            } else if(alignMethod == 3) {
-                alignOrientedReads3(
-                    orientedReadIds[0], orientedReadIds[1],
-                    matchScore, mismatchScore, gapScore,
-                    downsamplingFactor, bandExtend,
-                    alignment, alignmentInfo
-                );
+            if(storedAlignments.size() > 0) {
```
paoloczi, line 497: You could check using `storedAlignments.isOpen()` instead.

paoloczi:
The comments I made are cosmetic and optional, so I am merging the PR as is. You may want to consider making those changes at a later time.

bagashe:
Cool. I will make those changes some time today. Excited about a full assembly. Yay!

PR #118, opened 2020-04-17
Reuse code to compute total kmer count.

bagashe:
It threw me for a loop every time I saw it. So abstracting it away in an inline method next to the doc-block comment that explains the math.

### DIFF
--- a/src/Assembler.hpp
+++ b/src/Assembler.hpp
@@ -706,6 +706,9 @@ private:
     // kmerTable[i].isMarker == kmerTable[kmerTable[i].reverseComplementKmerId].isMarker
     MemoryMapped::Vector<KmerInfo> kmerTable;
     void checkKmersAreOpen() const;
+    inline size_t getKmerCount(size_t k) const {
+      return(1ULL << (2ULL*k));
+    };
 
 public:
     void accessKmers();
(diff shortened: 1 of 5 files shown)


paoloczi:
This is a good thing to do, of course. However it seems to me that this function would belong better in a static member function of template class `ShortBaseSequence`, because it returns the number of distinct `Kmer`'s of `k` bases, and `Kmer` is a `ShortBaseSequence`.

Also I would prefer if the return type was `uint64_t` instead of `size_t`. I am trying to move away from `size_t` because it has different sizes on Linux and MacOS, which sometimes causes conversion pain and compilation problems on MacOS.

Also for the future please avoid acronyms like DRY in submit messages. Other things that I prefer avoided in submit messages are contractions, abbreviations, and emoticons. Each of these things may have an obvious meaning to the submitter but not necessarily to the reader.

bagashe:
All good points. I need to be more mindful about using acronyms and assuming everyone knows what they mean :)

bagashe:
@paoloczi : I moved the method to Kmer.hpp/cpp. Moving it to ShortBaseSequence seemed weird since Kmer is a specific instantiation of it. Does this seem reasonable?

paoloczi:
As I suggested above, this function would belong better in a static member function of template class `ShortBaseSequence`, because it returns the number of distinct `Kmer`'s of `k` bases, and `Kmer` is a `ShortBaseSequence`. That would be more natural than creating a new namespace just for this purpose.

PR #126, opened 2020-05-04
Build on Ubuntu 20.04 LTS

bagashe:
Ubuntu 20.04 LTS ships with gcc v9.3.0, which throws a warning where we don't catch
exceptions by reference. Fixed three such occurrences.

On Ubuntu 20.04 LTS
`apt install cmake` installs version 3.16.3 (as of today).
`apt install python3` installs version 3.8.2 (as of today).

So I had to modify a cmake input file to deal with the changes in the way python is embedded
into an application. See
https://docs.python.org/3.8/whatsnew/3.8.html#debug-build-uses-the-same-abi-as-release-build for
more details.

Test Plan: Spun up an ec2 instance with an Ubuntu 20.04 LTS AMI. Applied these changes. Built
successfully and ran a small assembly. It works. Same for Ubuntu 18.04 LTS to verify that these
changes are backward compatible.

### DIFF
--- a/dynamicLibrary/CMakeLists.txt
+++ b/dynamicLibrary/CMakeLists.txt
@@ -74,9 +74,20 @@ set_target_properties(shastaDynamicLibrary PROPERTIES DEFINE_SYMBOL "")
 # Python 3 and pybind11.
 # I was not able to get find_package to work with pybind11 as installed by pip3.
 # The code below creates include paths for both Python 3 and pybind11.
+
+# Python 3.8 requires the `--embed` flag to be passed in.
+# find_python cmake module was introduced in v3.12, which isn't available for installation
+# via `apt` on Ubuntu 18.04 LTS. So I am not using it. 
+execute_process(COMMAND /usr/bin/python3-config --embed --libs OUTPUT_VARIABLE
+SHASTA_PYTHON_LIBRARIES RESULT_VARIABLE COMMAND_RESULT)
+if(${COMMAND_RESULT} EQUAL 0)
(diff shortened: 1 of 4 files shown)


dynamicLibrary/CMakeLists.txt
```diff
@@ -74,9 +74,20 @@ set_target_properties(shastaDynamicLibrary PROPERTIES DEFINE_SYMBOL "")
 # Python 3 and pybind11.
 # I was not able to get find_package to work with pybind11 as installed by pip3.
 # The code below creates include paths for both Python 3 and pybind11.
+
+# Python 3.8 requires the `--embed` flag to be passed in.
+# find_python cmake module was introduced in v3.12, which isn't available for installation
+# via `apt` on Ubuntu 18.04 LTS. So I am not using it. 
+execute_process(COMMAND /usr/bin/python3-config --embed --libs OUTPUT_VARIABLE
```
paoloczi, line 81: Would it be more straightforward to use `python3 --version`, then use its output to decide whether to add the `--embed` flag?

bagashe: Considered it. Decided against it because we don't know if this change is permanent. They might revert back for reasons we don't know of in a later version. If they do, then the conditional gets more complicated and we might have to maintain a white list of versions that require the `--embed` flag. It also made the code more complex. I couldn't do the recommended way because `find_python` does not work on the version of `cmake` available through `apt` on Ubuntu 18.04.

Eventually, as we drop support for older versions of Ubuntu, we can clean up this mess.

PR #131, opened 2020-05-17
Alignment configuration table revamp and high level style changes.

bagashe:
1. `exploreAlignment` & `computeAllAlignments` now have a description for each parameter. The description is pulled from `boost::program_options::options_description`, which is populated from the shasta config file.
2. We now use the actual config param name in the first column. This makes the table look quite similar to the `[Align]` section in the configuration file, thereby reducing the cognitive load of getting started with the HTTP server for debugging. 
3. Redid the style sheet to
- De-emphasize the table borders, so that viewer can focus on the data.
- Created a style for `input`, so that the component looks the same on all pages.

image

image

image

### DIFF
--- a/src/Assembler.hpp
+++ b/src/Assembler.hpp
@@ -2035,6 +2035,21 @@ public:
     void exploreMarkerGraphEdge(const vector<string>&, ostream&);
     void exploreMarkerCoverage(const vector<string>&, ostream&);
     void exploreMarkerGraphInducedAlignment(const vector<string>&, ostream&);
+    void renderEditableAlignmentConfig(
+        const int method,
+        const uint64_t maxSkip,
+        const uint64_t maxDrift,
+        const uint32_t maxMarkerFrequency,
+        const uint64_t minAlignedMarkerCount,
+        const double minAlignedFraction,
+        const uint64_t maxTrim,
+        const int matchScore,
+        const int mismatchScore,
+        const int gapScore,
+        const double downsamplingFactor,
+        const uint32_t bandExtend,
+        ostream& html
+    );
 #endif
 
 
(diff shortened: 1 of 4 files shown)


src/Assembler.hpp
```diff
@@ -2035,6 +2035,21 @@ class shasta::Assembler :
...
+        const uint64_t maxSkip,
+        const uint64_t maxDrift,
+        const uint32_t maxMarkerFrequency,
+        const uint64_t minAlignedMarkerCount,
+        const double minAlignedFraction,
+        const uint64_t maxTrim,
+        const int matchScore,
+        const int mismatchScore,
+        const int gapScore,
+        const double downsamplingFactor,
```
bagashe, line 2049: I was wondering if I could reuse `AssemblerOptions::AlignOptions` here, but turns out that I can't really forward declare an inner class :-)

I wondered if I could reuse `ComputeAllAlignmentsData` here, but decided against it in order to keep the compute and render code separate.

So I ended up with this method signature.

paoloczi: If you found it useful, it would be ok to `#include "AssemblerOptions.hpp"` in `Assembler.hpp`. Then you would be able to use `AlignOptions` there. Or conceivably even make this function a member of `AlignOptions`. Alternatively, moving `AlignOptions` out of `AssemblerOptions` would be fine too. Or you can leave it as is, perfectly fine.

I agree with not using `ComputeAllAlignmentsData`. I prefer those classes used to communicate with threads to be used by one function each. That makes it easier to make changes, when needed, without having to test other pieces of functionality.

paoloczi:
Nice - thank you!

PR #132, opened 2020-05-18
Shasta compiles with SeqAn on MacOS.

bagashe:
Test Plan: MacOS build plans succeed in Github actions. Looking at the logs, verified that it installed Seqan v2.4.0. There's no way to specify the exact version - the best I can do is `seqan@2`. It should be fine though because SeqAn 3 is out and I don't expect there to be many more 2.* releases.

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -29,7 +29,7 @@ jobs:
       run: |
         sudo apt install gzip
         gunzip tests/TinyTest.fasta.gz
-        shasta-build/shasta-Ubuntu-18.04/bin/shasta --input tests/TinyTest.fasta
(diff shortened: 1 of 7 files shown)


.github/workflows/Build.yml
```diff
@@ -67,7 +67,7 @@ jobs:
       run: |
         sudo apt install gzip
         gunzip tests/TinyTest.fasta.gz
-        shasta-build/shastaGpu-Ubuntu-18.04/bin/shastaGpu --input tests/TinyTest.fasta
+        shasta-build/shastaGpu-Ubuntu-18.04/bin/shastaGpu --Align.alignMethod 3 --input tests/TinyTest.fasta
```
paoloczi, line 70: The GPU code does not honor alignMethod, but it does not really matter.

PR #135, opened 2020-05-27
Clean up docs and Build files.

bagashe:
cmake is now configured to generate a portable binary by default. Cleaning up the docs to reflect that.

### DIFF
--- a/docs/BuildingFromSource.html
+++ b/docs/BuildingFromSource.html
@@ -167,43 +167,6 @@ build directory.
 
 
 
-<h3 id=Portability>Build portability</h3>
-<p>
-For best assembly performance, the build process described above
-creates binaries that use the full instruction set
-available on the build machine (via gcc option <code>-march=native</code>).
-This gives the best possible runtime performance on the build machine,
-but <b>creates a build that is not necessarily portable across machines</b>.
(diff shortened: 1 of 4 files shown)


docs/BuildingFromSource.html
```diff
@@ -170,39 +170,8 @@ <h3>Building smaller binaries</h3>
 
 <h3 id=Portability>Build portability</h3>
```
paoloczi, line 171: Perhaps remove this entire section? This is what people expect anyway. The section was necessary because the previous behavior was different from this common expectation.

docs/MakeRelease.html
```diff
@@ -116,7 +116,7 @@ <h2>Procedure used for release 0.1.0</h2>
 git clone https://github.com/chanzuckerberg/shasta.git
```
paoloczi, line 116: This entire section was the old procedure used for release 0.1.0 which is obsolete and no longer applies. It is time to remove this entire section.

paoloczi: Also, remove the previous header "Current procedure". This file is for our use only anyway and it is not linked anywhere in the rest of the documentation.

bagashe:
Will do

PR #137, opened 2020-05-28
Adding a configuration option to skip flagging of Palindromic reads.

bagashe:
Verfied that passing a boolean switch command line option `--Reads.palindromicReads.skipFlagging` skips that step.

### DIFF
--- a/docs/CommandLineOptions.html
+++ b/docs/CommandLineOptions.html
@@ -364,7 +364,7 @@ Can help performance, but only use it if you know you will not
 need to access the input files again soon.
 
 <tr id='Reads.palindromicReads.skipFlagging'>
-<td><code>--Reads.palindromicReads.skipFlaging</code><td class=centered><code>False</code><td>
+<td><code>--Reads.palindromicReads.skipFlagging</code><td class=centered><code>False</code><td>
 Skip flagging palindromic reads. Oxford Nanopore reads should be flagged for better results.
 
 <tr id='Reads.palindromicReads.maxSkip'>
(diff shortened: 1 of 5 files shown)


docs/CommandLineOptions.html
```diff
@@ -371,6 +371,10 @@ <h2>Options allowed on the command line and in the config file</h2>
 Can help performance, but only use it if you know you will not 
 need to access the input files again soon.
 
+<tr id='Reads.palindromicReads.skipFlagging'>
+<td><code>--Reads.palindromicReads.skipFlaging</code><td class=centered><code>False</code><td>
```
paoloczi, line 375: The option name is mispelled (missing a `g`).

src/AssemblerOptions.cpp
```diff
@@ -569,6 +574,7 @@ void AssemblerOptions::addConfigurableOptions()
 
 void AssemblerOptions::ReadsOptions::PalindromicReadOptions::write(ostream& s) const
 {
+    s << "palindromicReads.skipFlagging = " << skipFlagging << "\n";
```
paoloczi, line 577: Need to use `convertBoolToPythonString` as done for other Boolean switches. This is necessary so the configuration file stored in the assembly directory can be used by the Python API.

docs/CommandLineOptions.html
bagashe: Good catch. Will fix.

src/AssemblerOptions.cpp
bagashe: Ah. Didn't realize that. Will do.

PR #139, opened 2020-06-03
Name the Ubuntu-20-04 artifact shasta-Linux-Beta

bagashe:
...  to distinguish it from shasta-Linux which is built on Ubuntu 18.04.

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -64,10 +64,6 @@ jobs:
         ls -l ShastaRun/Assembly.fasta
     - uses: actions/upload-artifact@master
       with:
-        name: shasta-Linux
-        path: shasta-build/shasta-Ubuntu-18.04/bin/shasta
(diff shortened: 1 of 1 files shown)


bagashe:
Technically, there is nothing `Beta` about that portable binary, other than the fact that it runs on a more up-to-date version of the kernel and system libraries. I couldn't think of a better name. Open to suggestions.

paoloczi:
My inclination would be to call it `shasta-Linux` and eliminate the current `shasta-Linux` artifact built on Ubuntu 18.04. This would minimize user confusion, I think. And this way most users would be running the version built on Ubuntu 20.04, which is the right thing to do.

Those with an old Linux kernel would use `shasta-OldLinux` which continues to be built on Ubuntu 16.04. If somebody really needs the executable built on Ubuntu 18.04 they can always extract it from the tar file for Ubuntu 18.04, which continues to be available as another build artifact.

PR #141, opened 2020-06-04
Upgrade to SPOA 3.4.0 

bagashe:
SPOA 3.4.0 has the following relevant changes ...
1. It supports cpu-dispatching which allows run-time selection of available instruction sets on a processor. It takes away this decision from the hands of `gcc` which frankly, sucks at it, based on all experiments that I have run. The phase of Shasta that uses SPOA speeds up 4-5X with this upgrade. 

2. It fixes an error in clearing out a Graph data structure. This would open up the possibility of re-using the object in our loop, potentially reducing memory allocations and speeding things up.

Note that cpu-dispatching code does not compile on Ubuntu 16.04 LTS. So `oldLinux` shasta executables will be built with the portable version of SPOA. I have added a new cmake flag `-DBUILD_USE_SPOA_WITH_CPU_DISPATCH`, which defaults to `ON` because, its quite unlikely that anyone is going to build Shasta from source on such an old version of Linux and they should use the static executable in the release (or github actions) instead. This flag is set to `OFF` in `Build.yml` for building `shasta-OldLinux`.

Test Plan: Github actions.


CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ if(MACOS)
     option(BUILD_APPIMAGE "Build the AppImage." OFF)
 else(MACOS)
     # For Linux, build everything.
+    option(BUILD_USE_SPOA_WITH_CPU_DISPATCH "Link to SPOA library which does not use cpu dispatch" ON)
```
bagashe, line 32: Ugh. Will fix description.

bagashe:
Verified in Github actions logs that BUILD_USE_SPOA_WITH_CPU_DISPATCH is set to OFF for Ubuntu 16.04 and ON for Ubuntu 18.04 and 20.04.

PR #146, opened 2020-06-10
Minor changes to documentation.

bagashe:
I rewrote the QuickStart, BuildingFromSource and SupportedPlatforms documentation.

You can see it at https://bagashe.github.io/shasta/QuickStart.html. I wanted to avoid using hard-coded version names in the documentation. I also wanted to provide a list of commands that users can copy-paste and have it work without having to install any of those commands first. MacOS doesn't come with `wget` pre-installed. So I use `curl` instead. 

We do support building on macOS 10.14 and 10.15. So I have updated BuildingFromSource.html to reflect that.

### DIFF
--- a/docs/BuildingFromSource.html
+++ b/docs/BuildingFromSource.html
@@ -59,7 +59,7 @@ Even better, submit a Pull Request with the necessary changes.
 <pre>
 git clone https://github.com/chanzuckerberg/shasta.git
 </pre>
-<li>Install prerequisite packages
+<li>Install <a href="Prerequisites.html">prerequisite packages</a>
 <pre>
 sudo shasta/scripts/InstallPrerequisites-Ubuntu.sh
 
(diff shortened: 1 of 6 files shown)


bagashe:
Check out https://bagashe.github.io/shasta/index.html for how it looks.

paoloczi:
This is very nice - thank you. Just one last question. Several titles begin with a verb in its root form like "Run" or "Explore". This was fine before because these titles were under the "How to" header - so "How to run", "How to explore". But now that the "How to" header is gone (and I don't miss it a bit), would it sound better if we used verbs in the [_-ing_](https://en.wikipedia.org/wiki/-ing#Uses) form - "Running" or "Exploring"? I will leave it for you to decide which is better.

bagashe:
> This is very nice - thank you. Just one last question. Several titles begin with a verb in its root form like "Run" or "Explore". This was fine before because these titles were under the "How to" header - so "How to run", "How to explore". But now that the "How to" header is gone (and I don't miss it a bit), would it sound better if we used verbs in the [_-ing_](https://en.wikipedia.org/wiki/-ing#Uses) form - "Running" or "Exploring"? I will leave it for you to decide which is better.

I agree. Entries in Table of contents should be consistent. I will update.

PR #155, opened 2020-06-17
Log peak memory usage.

bagashe:
If `/proc/self/status` does not exist (macOS) then it is not captured. This is fine since our goal is to use this to minimize peak memory usage for big assemblies (Linux). 

Test Plan: Looked at the stdout log for test assemblies on Github actions. Saw the value on Ubuntu, but not on macOS.

### DIFF
--- a/src/Assembler.cpp
+++ b/src/Assembler.cpp
@@ -150,3 +150,7 @@ void Assembler::storeAssemblyTime(
     assemblerInfo->averageCpuUtilization = averageCpuUtilization;
 }
 
+void Assembler::storePeakMemoryUsage(const string& peakMemoryUsage) {
+    assemblerInfo->peakMemoryUsage = peakMemoryUsage;
+}
+
--- a/src/Assembler.cpp
+++ b/src/Assembler.cpp
@@ -150,7 +150,7 @@ void Assembler::storeAssemblyTime(
     assemblerInfo->averageCpuUtilization = averageCpuUtilization;
 }
 
-void Assembler::storePeakMemoryUsage(const string& peakMemoryUsage) {
+void Assembler::storePeakMemoryUsage(const uint64_t peakMemoryUsage) {
     assemblerInfo->peakMemoryUsage = peakMemoryUsage;
 }
 
--- a/src/Assembler.cpp
+++ b/src/Assembler.cpp
@@ -150,7 +150,7 @@ void Assembler::storeAssemblyTime(
     assemblerInfo->averageCpuUtilization = averageCpuUtilization;
 }
 
-void Assembler::storePeakMemoryUsage(const uint64_t peakMemoryUsage) {
+void Assembler::storePeakMemoryUsage(uint64_t peakMemoryUsage) {
     assemblerInfo->peakMemoryUsage = peakMemoryUsage;
 }
 
--- a/src/Assembler.hpp
+++ b/src/Assembler.hpp
@@ -128,6 +128,9 @@ public:
     double averageCpuUtilization;
     uint64_t peakMemoryUsage = 0ULL;
 
+    inline string peakMemoryUsageForSummaryStats() {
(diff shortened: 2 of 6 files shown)


src/Assembler.hpp
```diff
@@ -126,6 +126,7 @@ class shasta::AssemblerInfo {
     // Performance statistics.
     double assemblyElapsedTimeSeconds = 0.;
     double averageCpuUtilization;
+    string peakMemoryUsage = "";
```
paoloczi, line 129: Perhaps store it as a uint64_t, in bytes? That would make it easier to do computations with it if we want to - for example report the number of bytes used per input base.

srcMain/main.cpp
```diff
@@ -839,6 +839,24 @@ void shasta::main::assemble(
         assembler.writeOrientedReadsByAssemblyGraphEdge();
     }
 
+    string peakMemoryUsage;
```
paoloczi, line 842: Make this block into a separate function `getPeakMemoryUsage`? It could go in `platformDependent.hpp` and `platformDependent.cpp` or in its own pair of `.hpp`, `.cpp` files.

src/Assembler.cpp
```diff
@@ -150,3 +150,7 @@ void Assembler::storeAssemblyTime(
     assemblerInfo->averageCpuUtilization = averageCpuUtilization;
 }
 
+void Assembler::storePeakMemoryUsage(const uint64_t peakMemoryUsage) {
```
paoloczi, line 153: I would avoid arguments that are `const` and passed by value. There is some discussion on the web of why this is a questionable thing to do.

src/AssemblerHttpServer.cpp
```diff
@@ -839,6 +839,8 @@ void Assembler::writeAssemblySummaryBody(ostream& html)
         "<td class=right>" << assemblerInfo->assemblyElapsedTimeSeconds/3600. <<
         "<tr><td>Average CPU utilization"
         "<td class=right>" << assemblerInfo->averageCpuUtilization <<
+        "<tr><td>Peak Memory utilization (bytes)"
+        "<td class=right>" << assemblerInfo->peakMemoryUsage <<
```
paoloczi, line 843: For macOS, `peakmemoryUsage` will be `0`. Write `Not determined` rather than `0`, or equivalent message to that effect. Similar comment below for `json` output.

src/platformDependent.cpp
```diff
@@ -35,3 +36,28 @@ std::string shasta::timeoutCommand()
...
+            }
+            char* end;
+            peakMemoryUsage = std::strtoull(line.c_str() + pos, &end, 10);
+            // Convert from kB to bytes.
+            peakMemoryUsage *= 1024;
+            break;
+        }
+    }
+#endif
+    return(peakMemoryUsage);
```
paoloczi, line 62: The parentheses are unnecessary and are generally not used in `return` statements in Shasta.

src/Assembler.cpp
bagashe: It doesn't matter in this case. In general, any `const` directive is good for enforcing that someone doesn't accidentally modify the value (even if it is a copy).

src/AssemblerHttpServer.cpp
bagashe: Sure.

src/Assembler.cpp
paoloczi: Of course. But the rationale for avoiding this is that the fact that the function does not modify the value of its copy is internal business of the function and should not be visible in its declaration, accessible to the caller. If you want to make it clear that the function is not modifying its copy you could use a pattern like this:

```
void f(T x)
{
    const T& xConst = x;
}
```

There is some discussion of this [here](http://www.cplusplus.com/forum/general/135731/), where somebody also suggests the possibility of using the `const` qualifier in the definition of the function but not in the declaration. That would avoid the extra complexity of the above pattern. 

But this gets a bit into philosophical differences, so feel free to write it as you please.

PR #161, opened 2020-06-23
Multi-read graph display, with more layout options

rlorigro:
Added ability to display graph starting from multiple reads and use multiple layout methods. Also removed unused display option for containment edges.

### DIFF
--- a/src/Assembler.hpp
+++ b/src/Assembler.hpp
@@ -1067,12 +1067,23 @@ private:
     void flagChimericReadsThreadFunction(size_t threadId);
 
 
-
     // Create a local subgraph of the global read graph,
     // starting at a given vertex and extending out to a specified
     // distance (number of edges).
     bool createLocalReadGraph(
-        OrientedReadId start,
+            OrientedReadId start,
+            uint32_t maxDistance,   // How far to go from starting oriented read.
+            bool allowChimericReads,
+            bool allowCrossStrandEdges,
+            size_t maxTrim,         // Used to define containment.
+            double timeout,         // Or 0 for no timeout.
+            LocalReadGraph&);
+
+    // Create a local subgraph of the global read graph,
+    // starting at any number of  given vertexes and extending out to a specified
+    // distance (number of edges).
+    bool createLocalReadGraph(
+        vector<OrientedReadId> starts,
         uint32_t maxDistance,   // How far to go from starting oriented read.
         bool allowChimericReads,
         bool allowCrossStrandEdges,
--- a/src/Assembler.hpp
+++ b/src/Assembler.hpp
@@ -1083,7 +1083,7 @@ private:
     // starting at any number of  given vertexes and extending out to a specified
     // distance (number of edges).
     bool createLocalReadGraph(
-        vector<OrientedReadId> starts,
+        vector<OrientedReadId>& starts,
         uint32_t maxDistance,   // How far to go from starting oriented read.
         bool allowChimericReads,
         bool allowCrossStrandEdges,
--- a/src/AssemblerHttpServer-ReadGraph.cpp
+++ b/src/AssemblerHttpServer-ReadGraph.cpp
@@ -31,18 +31,53 @@ void Assembler::exploreReadGraph(
 }
 
 
+bool parseCommaSeparatedReadIDs(string& commaSeparatedReadIDs, vector<OrientedReadId>& readIds, ostream& html){
+    readIds.clear();
+    string token;
+
+    for (auto& c: commaSeparatedReadIDs){
+        if (c == ','){
+            try {
+                OrientedReadId readID(token);
+                readIds.emplace_back(readID);
+                token.clear();
+            }
+            catch(exception& e){
+                html << "<p>Invalid read id or read strand: '" << token << "'</p>";
+                html << "<p>" << e.what() << "</p>";
+                return false;
+            }
+        }
+        else{
+            token += c;
+        }
+    }
+
+    // Place final token which may not have a comma after it
+    if (not token.empty()) {
+        try {
+            OrientedReadId readID(token);
+            readIds.emplace_back(readID);
+        }
+        catch(exception& e){
+            html << "<p>Invalid read id or read strand: '" << token << "'</p>";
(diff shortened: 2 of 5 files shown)


src/Assembler.hpp
```diff
@@ -1067,12 +1067,23 @@ class shasta::Assembler :
...
+            bool allowCrossStrandEdges,
+            size_t maxTrim,         // Used to define containment.
+            double timeout,         // Or 0 for no timeout.
+            LocalReadGraph&);
+
+    // Create a local subgraph of the global read graph,
+    // starting at any number of  given vertexes and extending out to a specified
+    // distance (number of edges).
+    bool createLocalReadGraph(
+        vector<OrientedReadId> starts,
```
bagashe, line 1086: This will `copy` the vector. Pass it as a reference instead? - `const vector<OrientedReadId>& starts`

rlorigro: Yes, good catch, thanks

src/AssemblerHttpServer-ReadGraph.cpp
```diff
@@ -31,18 +31,53 @@ void Assembler::exploreReadGraph(
 }
 
 
+bool parseCommaSeparatedReadIDs(string& commaSeparatedReadIDs, vector<OrientedReadId>& readIds, ostream& html){
```
paoloczi, line 34: Shasta uses naming conventions where, in a name containing an acronym, only the first letter is capitalized. So in the name of this function and the first argument `ID` should become `Id`. This is also consistent with `OrientedReadId` and `readIds`.

src/AssemblerHttpServer-ReadGraph.cpp
```diff
@@ -31,18 +31,53 @@ void Assembler::exploreReadGraph(
...
+    readIds.clear();
+    string token;
+
+    for (auto& c: commaSeparatedReadIDs){
+        if (c == ','){
+            try {
+                OrientedReadId readID(token);
+                readIds.emplace_back(readID);
```
paoloczi, line 42: I would not bother with `emplace_back` here and use the more standard `push_back` instead. Unless I misunderstand something, it provides no benefits for this code where you are just dealing with integers.

src/AssemblerHttpServer-ReadGraph.cpp
paoloczi: The same comment also in a few more places below.

src/AssemblerReadGraph.cpp
```diff
@@ -161,116 +161,142 @@ void Assembler::checkReadGraphIsOpen()
...
-    q.push(start);
+        // Initialize a BFS starting at the start vertex.
+        std::queue<OrientedReadId> q;
+        q.push(start);
 
 
 
-    // Do the BFS.
-    while(!q.empty()) {
+        // Do the BFS.
```
paoloczi, line 221: Wouldn't it be better to do a single BFS, initializing the queue with all of the start vertices?

src/AssemblerHttpServer-ReadGraph.cpp
rlorigro: As written, I dont think it would provide any benefit. If instead I did:
`readIds.emplace_back(token);`

It would avoid copying or moving the `readId` object. Would you be ok with this implementation?

paoloczi: No, the compiler would optimize all of that away anyway. I think `emplace_back` is meant for situations where there is a significant cost in object construction, and/or object construction is complex enough that the compiler cannot do a good job inlining it.

And more in general, in areas of the code that are not important for performance, we should avoid optimizations that make the code less readable.

src/AssemblerReadGraph.cpp
```diff
@@ -207,32 +235,32 @@ bool Assembler::createLocalReadGraph(
         const uint32_t distance1 = distance0 + 1;
 
         // Loop over edges of the global read graph involving this vertex.
-        for(const uint64_t i: readGraph.connectivity[orientedReadId0.getValue()]) {
+        for (const uint64_t i: readGraph.connectivity[orientedReadId0.getValue()]) {
             SHASTA_ASSERT(i < readGraph.edges.size());
-            const ReadGraphEdge& globalEdge = readGraph.edges[i];
+            const ReadGraphEdge &globalEdge = readGraph.edges[i];
```
paoloczi, line 240: For consistency with the Shasta standard style and common C++ usage, please put this back as it was, that is, change to the following (space after the `&`, not before):

```
const ReadGraphEdge& globalEdge
```
There is at least one more place below (when defining `alignment`) where a similar change 
 back to the original code is needed.

src/AssemblerReadGraph.cpp
```diff
@@ -207,32 +235,32 @@ bool Assembler::createLocalReadGraph(
...
             }
 
             // Get alignment information.
-            const AlignmentData& alignment = alignmentData[globalEdge.alignmentId];
+            const AlignmentData &alignment = alignmentData[globalEdge.alignmentId];
             OrientedReadId alignmentOrientedReadId0(alignment.readIds[0], 0);
             OrientedReadId alignmentOrientedReadId1(alignment.readIds[1], alignment.isSameStrand ? 0 : 1);
             AlignmentInfo alignmentInfo = alignment.info;
-            if(alignmentOrientedReadId0.getReadId() != orientedReadId0.getReadId()) {
+            if (alignmentOrientedReadId0.getReadId() != orientedReadId0.getReadId()) {
```
paoloczi, line 259: I am not necessarily against a space  preceding parentheses after all `if`, but unrelated formatting changes make PRs harder to review. No need to change them back.

src/AssemblerReadGraph.cpp
rlorigro: Yes, my mistake, I actually prefer the `&` to be part of the type as well. It must have been swapped by my IDE

src/AssemblerReadGraph.cpp
rlorigro: I think this is another case of the IDE changing things

PR #164, opened 2020-07-01
Document the Kernel too old error and how to deal with it.

bagashe:


docs/QuickStart.html
```diff
@@ -30,6 +30,13 @@ <h2 id="QuickStartLinux">Linux</h2>
 ./shasta-Linux-0.5.1 --input input.fasta
 </pre>
 
+If the above command fails with <code>FATAL: kernel too old</code>, then you should retry
```
paoloczi, line 33: I would prefer to keep the Quick Start documentation as short as possible, to avoid creating the impression that the code is hard to use. Perhaps put this information under "Running an assembly"? 

The Quick start documentation already says "This will work on most current 64 bit Linux distributions", so we are covered. We don't need to cover the unusual case there. But I like the text.

PR #165, opened 2020-07-02
Added PeakFinder and integrated into Assembler

rlorigro:
This adds a method for dynamically selecting a Marker.minCoverage cutoff, using the distribution of coverages from the marker disjoint set object during runtime. 

So far it passes all test cases I could generate, but I am going to spend a bit more time evaluating the no-peak cases, so it shouldn't be merged until then.

### DIFF
--- a/src/AssemblerMarkerGraph.cpp
+++ b/src/AssemblerMarkerGraph.cpp
@@ -230,6 +230,7 @@ void Assembler::createMarkerGraphVertices(
             shasta::PeakFinder p;
             p.findPeaks(histogram);
             minCoverage = p.findXCutoff(histogram);
+            cout << "Marker.minCoverage inferred during run time: " << minCoverage << '\n';
         }
     }
 
--- a/src/AssemblerMarkerGraph.cpp
+++ b/src/AssemblerMarkerGraph.cpp
@@ -3,6 +3,7 @@
 #include "AlignmentGraph.hpp"
 #include "ConsensusCaller.hpp"
 #include "compressAlignment.hpp"
+#include "PeakFinder.hpp"
 #ifdef SHASTA_HTTP_SERVER
 #include "LocalMarkerGraph.hpp"
 #endif
@@ -223,6 +224,12 @@ void Assembler::createMarkerGraphVertices(
             if(frequency) {
                 csv << coverage << "," << frequency << "\n";
             }
+        }
+
+        if (minCoverage == 0) {
+            shasta::PeakFinder p;
+            p.findPeaks(histogram);
+            minCoverage = p.findXCutoff(histogram);
         }
     }
 
--- a/src/AssemblerMarkerGraph.cpp
+++ b/src/AssemblerMarkerGraph.cpp
@@ -227,10 +227,15 @@ void Assembler::createMarkerGraphVertices(
         }
 
         if (minCoverage == 0) {
-            shasta::PeakFinder p;
-            p.findPeaks(histogram);
-            minCoverage = p.findXCutoff(histogram);
-            cout << "Marker.minCoverage inferred during run time: " << minCoverage << '\n';
+            try {
+                shasta::PeakFinder p;
+                p.findPeaks(histogram);
+                minCoverage = p.findXCutoff(histogram);
+                cout << "MarkerGraph.minCoverage inferred during run time: " << minCoverage << '\n';
+            }
+            catch (PeakFinderException){
+                throw runtime_error("ERROR: No significant cutoff found in distribution");
+            }
         }
     }
 
--- a/src/PeakFinder.cpp
+++ b/src/PeakFinder.cpp
@@ -2,8 +2,6 @@
 #include <algorithm>
 
 using namespace shasta;
-using std::sort;
-using std::cerr;
 
 
 PeakFinder::Peak::Peak(uint64_t start):
--- a/src/PeakFinder.cpp
+++ b/src/PeakFinder.cpp
@@ -6,6 +6,16 @@ using std::sort;
 using std::cerr;
 
 
+shasta::Peak::Peak(uint64_t start):
+        start(start),
+        stop(0),
+        left(start),
+        right(start),
+        isMerged(false),
+        persistence(0)
+{}
+
+
 void PeakFinder::findPeaks(vector<uint64_t>& y){
     size_t xSize = y.size();
 
--- a/src/PeakFinder.cpp
+++ b/src/PeakFinder.cpp
@@ -153,6 +153,11 @@ uint64_t PeakFinder::calculateArea(const vector<uint64_t>& y, uint64_t xMin, uin
 
 
 uint64_t PeakFinder::findXCutoff(const vector<uint64_t>& y){
+    // First check that there is at least one peak (beyond the expected error peak at x=1)
+    if (peaks.size() < 2) {
+        throw PeakFinderException();
+    }
+
     sortByPersistence();
 
     // find the total AUC
--- a/src/PeakFinder.cpp
+++ b/src/PeakFinder.cpp
@@ -16,7 +16,7 @@ shasta::Peak::Peak(uint64_t start):
 {}
 
 
-void PeakFinder::findPeaks(vector<uint64_t>& y){
+void PeakFinder::findPeaks(const vector<uint64_t>& y){
     size_t xSize = y.size();
 
     // Create a vector of values, in which each value can be used to fetch the parent peak for that position in the
@@ -140,7 +140,7 @@ void PeakFinder::sortByPersistence(){
 }
 
 
-uint64_t PeakFinder::calculateArea(vector<uint64_t>& y, uint64_t x_min, uint64_t x_max){
+uint64_t PeakFinder::calculateArea(const vector<uint64_t>& y, uint64_t x_min, uint64_t x_max){
     uint64_t total = 0;
 
     for (size_t i=x_min; i<=x_max; i++){
@@ -151,7 +151,7 @@ uint64_t PeakFinder::calculateArea(vector<uint64_t>& y, uint64_t x_min, uint64_t
 }
 
 
-uint64_t PeakFinder::findXCutoff(vector<uint64_t>& y){
+uint64_t PeakFinder::findXCutoff(const vector<uint64_t>& y){
     sortByPersistence();
 
     // find the total AUC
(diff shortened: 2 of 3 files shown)


src/PeakFinder.hpp
```diff
@@ -0,0 +1,86 @@
...
+    /// Attributes ///
+
+    // This vector will be filled with peaks during iteration
+    vector<Peak> peaks;
+
+
+    /// Methods ///
+
+    // Iterate the y values and identify all the peaks/valleys
+    void findPeaks(vector<uint64_t>& y);
```
bagashe, line 77: I skimmed the implementation. Looks like this can be a `const vector<uint64_t>& y`.

src/PeakFinder.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+    void sortByPersistence();
+    uint64_t findXCutoff(vector<uint64_t>& y);
```
bagashe, line 80: Same here.

src/PeakFinder.hpp
```diff
@@ -0,0 +1,86 @@
...
+    uint64_t calculateArea(vector<uint64_t>& y, uint64_t x_min, uint64_t x_max);
```
bagashe, line 81: And here.

src/PeakFinder.hpp
```diff
@@ -0,0 +1,86 @@
...
+shasta::Peak::Peak(uint64_t start){
+    this->left = start;
+    this->right = start;
+    this->start = start;
+    this->stop = 0;
+    this->isMerged = false;
+}
+
+
+class shasta::PeakFinder {
```
bagashe, line 66: I can guess what `PeakFinder` is trying to do. What would really help someone new looking at this code later on, is if the basic idea and computation were documented in some form. Plain text works. Since it's a histogram, I bet you could use some simple illustrations as well using horizontal visualization. Something like ...

```
0 => ###
1 => ######
2 => #######################
3 => ####
4 => ######
...
// In this case PeakFinder would <insert explanation here>
```

It would also help me review the actual logic and validate whether the code is doing what's intended.

src/AssemblerMarkerGraph.cpp
```diff
@@ -224,6 +225,12 @@ void Assembler::createMarkerGraphVertices(
                 csv << coverage << "," << frequency << "\n";
             }
         }
+
+        if (minCoverage == 0) {
```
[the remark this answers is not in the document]
paoloczi: No. We decided `--MarkerGraph.minCoverage 0` will mean "automatic" and will trigger this code. @rlorigro and I decided that I will make the changes in the documentation and the option description.

src/AssemblerMarkerGraph.cpp
```diff
@@ -224,6 +225,12 @@ void Assembler::createMarkerGraphVertices(
...
+            shasta::PeakFinder p;
+            p.findPeaks(histogram);
+            minCoverage = p.findXCutoff(histogram);
+        }
```
paoloczi, line 233: We need a message writing out the `minCoverage` value that was chosen.

src/PeakFinder.hpp
```diff
@@ -0,0 +1,86 @@
...
+#define SHASTA_PEAKFINDER_HPP
+
+
+#include <experimental/filesystem>
+#include <stdexcept>
+#include <iostream>
+#include <utility>
+#include <vector>
+
+using std::experimental::filesystem::create_directories;
```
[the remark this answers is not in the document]
paoloczi: If this is somehow needed: we should not use `experimental/filesystem` because it will create portability issues. Instead, if necessary use the functions defined in namespace `shasta/filesystem` (see `shasta/src/filesystem.hpp`). If the functionality you need is missing, add it there. But I looks like none of this is actually needed, like @bagashe points out.

src/PeakFinder.hpp
```diff
@@ -0,0 +1,86 @@
+#ifndef SHASTA_PEAKFINDER_HPP
+#define SHASTA_PEAKFINDER_HPP
+
+
+#include <experimental/filesystem>
+#include <stdexcept>
+#include <iostream>
+#include <utility>
```
paoloczi, line 8: Don't `#include` `<utility>` and `<vector>`. Include the Shasta versions `"utility.hpp"` and `"vector.hpp"` instead, which also inject some `std` names in the `shasta` namespace.

src/PeakFinder.hpp
```diff
@@ -0,0 +1,86 @@
...
+using std::cout;
+
+namespace shasta {
+    class Peak;
+    class PeakFinder;
+}
+
+class shasta::Peak {
+public:
+    /// Attributes ///
```
paoloczi, line 27: Perhaps make most of this `private`?

paoloczi: Or, perhaps better, make `Peak` a private subclass of `PeakFinder`, since it does not to be used outside not it?

src/PeakFinder.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+    // Persistence is the height from the top of the peak to its base, where it was merged with another peak
+    uint64_t persistence;
+
+    /// Methods ///
+    Peak(uint64_t start);
+};
+
+
+ostream& operator<<(ostream& o, shasta::Peak& peak){
```
paoloczi, line 50: As written, this function is in the global namespace and I prefer to avoid that. It should go in the `shasta` namespace. See the end of `ReadId.hpp` for an example of how to do that.

src/PeakFinder.hpp
```diff
@@ -0,0 +1,86 @@
...
+    o << peak.start << ',' << peak.stop << ',' << peak.isMerged << ',' << peak.left << ',' << peak.right << ',' << peak.persistence;
+
+    return o;
+}
+
+
+shasta::Peak::Peak(uint64_t start){
```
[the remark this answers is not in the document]
paoloczi: Yes, please use the initializer list syntax. Note that there is no ambiguity between the arguments and the data members, so you can use syntax like `start(start)`.

Regarding putting this in the `hpp` versus `cpp` file I am neutral.

src/PeakFinder.hpp
paoloczi: @rlorigro yes, please do that. You could also include that link to the description of the algorithm you gave me the other day.

src/PeakFinder.hpp
```diff
@@ -0,0 +1,86 @@
+#ifndef SHASTA_PEAKFINDER_HPP
```
paoloczi, line 1: Consistently with other Shasta `.hpp` files, please add an underscore after `PEAK`, as we use underscores to separate words in include guards.

src/PeakFinder.cpp
```diff
@@ -0,0 +1,170 @@
...
+    vector<int64_t> peakIndex(xSize, -1);
+
+    // Create a vector of x values which point to y values
+    vector<size_t> indexes(xSize, 0);
+    for (size_t i=0; i<indexes.size() ; i++) {
+        indexes[i] = i;
+    }
+
+    // Sort the x values based on their corresponding y
+    sort(indexes.begin(), indexes.end(),
```
paoloczi, line 23: I find these lambda constructs unclear, but I can see that in this case the alternative would introduce additional complexity.

src/PeakFinder.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+    // Second most prominent peak
+    Peak& p = peaks[1];
+
+    // Find the AUC inside the bounds of the peak (from y=0 and up)
+    uint64_t peakArea = calculateArea(y, p.left, p.right);
+
+    double percentArea = 100 * (double(peakArea) / double(total_area));
+
+    uint64_t x_cutoff;
```
paoloczi, line 158: Here and in a few more places there are some symbols left with names that use underscores. In Shasta, underscores are only used in uppercase names (for macros and include guards).

src/PeakFinder.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+    // Check if second most prominent peak is reasonable size (not a false peak)
+    if (percentArea > 1){   //TODO: readjust this proportion for histograms with 0=0 frequency
```
paoloczi, line 161: Remove the comment if this is no longer needed. Or, if necessary, expand the comment to a multiple line comment.

src/PeakFinder.hpp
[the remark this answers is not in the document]
rlorigro: Agh this is a vestige of the implementation I wrote separately for testing, I will remove it.

src/PeakFinder.hpp
rlorigro: I could imagine a use case outside of shasta where you would want to do things with a sorted list of Peak objects, but I agree in this context it doesn't apply

src/PeakFinder.cpp
```diff
@@ -0,0 +1,170 @@
...
+        x_cutoff = p.left;
+    }
+    else{
+        throw runtime_error("ERROR: No significant cutoff found in marker frequency histogram");
```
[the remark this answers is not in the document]
rlorigro: I resolved this using a standalone class based on our separate discussion, let me know if a nested class is preferred

src/AssemblerMarkerGraph.cpp
```diff
@@ -224,6 +225,18 @@ void Assembler::createMarkerGraphVertices(
...
+
+        if (minCoverage == 0) {
+            try {
+                shasta::PeakFinder p;
+                p.findPeaks(histogram);
+                minCoverage = p.findXCutoff(histogram);
+                cout << "MarkerGraph.minCoverage inferred during run time: " << minCoverage << '\n';
+            }
+            catch (PeakFinderException){
+                throw runtime_error("ERROR: No significant cutoff found in distribution");
```
paoloczi, line 237: When I make the changes in the assembler options description I will modify the message to be more explanatory. You can leave it as is for now.

src/PeakFinder.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+***********************************************************************************************************************/
+
+
+#include "stdexcept.hpp"
+#include "iostream.hpp"
+#include "utility.hpp"
+#include "vector.hpp"
+
+using std::runtime_error;
```
paoloczi, line 58: All these `using` are no longer needed because these names are injected into the `shasta` namespace by the above `#include`.

paoloczi:
Getting close! I you just remove those `using` I will merge it, then I will add an explanation in the description of `--MarkerGraph.minCoverage` in the help message and the documentation, and expand the error message. This is a nice piece of work that simplifies using Shasta, particularly at low coverage.

PR #170, opened 2020-07-22
New page for browser: Assess alignments

rlorigro:
This is the first draft of a new page in the assembly browser which allows automated sampling of reads for aligning one-to-all. Bulk sampling enables stats on `alignedFraction` and `alignedMarkerCount` to be collected efficiently. In addition, the ratio of stored:found alignments is computed.

In the future I would like to add sampling from dead ends as an option, and include more data about alignments overhangs found in stored vs computed alignments. This is also a first pass at deciding whether alignments thresholds can be automatically determined at run time.

Its not clear to me where the merge conflict is coming from. I can work that out if needed.

Some example screenshots below:

image
image
image

### DIFF
--- a/src/AssemblerHttpServer-Alignments.cpp
+++ b/src/AssemblerHttpServer-Alignments.cpp
@@ -15,8 +15,6 @@ using namespace shasta;
 #include <boost/uuid/uuid.hpp>
 #include <boost/uuid/uuid_generators.hpp>
 #include <boost/uuid/uuid_io.hpp>
-#include <boost/accumulators/accumulators.hpp>
-#include <boost/accumulators/statistics.hpp>
 
 // Seqan
 #ifdef SHASTA_HTTP_SERVER
@@ -25,17 +23,6 @@ using namespace shasta;
 
 #ifdef SHASTA_HTTP_SERVER
 
-using namespace boost::accumulators;
-
-
-typedef boost::iterator_range< std::vector <std::pair <double, double> >::iterator> densityIterator;
-
-typedef accumulator_set <double, features
-        <tag::density,
-        tag::count,
-        tag::min,
-        tag::max> >
-        densityAccumulator;
 
 
 void Assembler::exploreAlignments(
@@ -1238,45 +1225,6 @@ void Assembler::sampleReads(vector<OrientedReadId>& sample, uint64_t n, uint64_t
 }
 
 
-void writeDensityAccumulatorToHtml(ostream& html, densityAccumulator& accumulator, uint64_t sizePx){
-    html << "<table style='margin-top: 1em; margin-bottom: 1em'>";
-    html << "<tr>"
-            "<th class='centered'>Left edge"
-            "<th class='centered'>Right edge"
-            "<th class='centered'>Count"
-            "<th class='centered'>Density"
-            "<th class='centered'>Plot";
-
-    double maxCount = boost::accumulators::max(accumulator);
-
-    // Determine the scaling factor for the histogram bars
-    double scale = double(sizePx)/double(maxCount);
-
-    // Initialize an iterator for the "density" portion of the accumulator
-    densityIterator histogram = density(accumulator);
-
-    // Find out the bin size from the first 2 entries
-    double binSize = histogram[1].first - histogram[0].first;
-
-    for (size_t i=0; i<histogram.size(); i++){
-        const double leftBound = histogram[i].first;
-        const double rightBound = leftBound + binSize;
-        const double y = histogram[i].second;
-
-        html << std::fixed << std::setprecision(2) <<
-            "<tr>"
-            "<td class=centered>" << leftBound <<
-            "<td class=centered>" << rightBound <<
-            "<td class=centered>" << y*maxCount <<
-            "<td class=centered>" << y <<
-            "<td>"
-            "<div class=sketch title='alignedFractionHistogram' style='display:inline-block;margin:0px;padding:0px;"
-            "background-color:blue;height:6px;width:" << double(y)*maxCount*scale << "px;'></div>";
-    }
-    html << "</table>";
-}
-
-
 // Compute alignments on an oriented read against
 // all other oriented reads, using a sampling of reads.
 // Display some useful stats for these alignments.
@@ -1391,22 +1339,9 @@ void Assembler::assessAlignments(
     }
 
     // Initialize histograms
-//    IterativeHistogram alignedFractionHistogram(0,1,20);
-//    IterativeHistogram markerCountHistogram(0,3000,100);
-//    IterativeHistogram nAlignmentsHistogram(0,200,20);
-
-    densityAccumulator alignedFractionAccumulator(
-            tag::density::cache_size=100,
-            tag::density::num_bins=100);
-
-    densityAccumulator markerCountAccumulator(
-            tag::density::cache_size=100,
-            tag::density::num_bins=100);
-
-    densityAccumulator nAlignmentsAccumulator(
-            tag::density::cache_size=100,
-            tag::density::num_bins=100);
-
+    IterativeHistogram alignedFractionHistogram(0,1,20);
+    IterativeHistogram markerCountHistogram(0,3000,100);
+    IterativeHistogram nAlignmentsHistogram(0,200,20);
 
     vector<pair<OrientedReadId, AlignmentInfo> > allAlignments;
     vector<pair<OrientedReadId, AlignmentInfo> > allStoredAlignments;
@@ -1473,18 +1408,15 @@ void Assembler::assessAlignments(
         for (auto& a: alignments){
             allAlignments.push_back(a);
         }
-//        nAlignmentsHistogram.update(double(alignments.size()));
-        alignedFractionAccumulator(double(alignments.size()));
+        nAlignmentsHistogram.update(double(alignments.size()));
     }
 
     for (auto& item: allAlignments){
         auto alignment = item.second;
 
         // Increment histograms
-//        markerCountHistogram.update(alignment.markerCount);
-        markerCountAccumulator(alignment.markerCount);
-//        alignedFractionHistogram.update(alignment.minAlignedFraction());
-        nAlignmentsAccumulator(alignment.minAlignedFraction());
+        markerCountHistogram.update(alignment.markerCount);
+        alignedFractionHistogram.update(alignment.minAlignedFraction());
     }
 
     // Pixel width of histogram display
@@ -1492,14 +1424,11 @@ void Assembler::assessAlignments(
 
     html << "<br><br>";
     html << "<br><strong>Marker Count Distribution</strong>";
-//    markerCountHistogram.writeToHtml(html, histogramSize);
-    writeDensityAccumulatorToHtml(html, markerCountAccumulator, histogramSize);
+    markerCountHistogram.writeToHtml(html, histogramSize);
     html << "<br><strong>Aligned Fraction Distribution</strong>";
-//    alignedFractionHistogram.writeToHtml(html, histogramSize);
-    writeDensityAccumulatorToHtml(html, alignedFractionAccumulator, histogramSize);
+    alignedFractionHistogram.writeToHtml(html, histogramSize);
     html << "<br><strong>Alignment Quantity Distribution</strong>";
-//    nAlignmentsHistogram.writeToHtml(html, histogramSize);
-    writeDensityAccumulatorToHtml(html, nAlignmentsAccumulator, histogramSize);
+    nAlignmentsHistogram.writeToHtml(html, histogramSize);
     html << "<br><strong>Ratio of stored to found alignments</strong>";
     html << "<br>" << double(allStoredAlignments.size())/double(allAlignments.size());
     html << "<br";
--- a/src/AssemblerHttpServer-Alignments.cpp
+++ b/src/AssemblerHttpServer-Alignments.cpp
@@ -1427,7 +1427,7 @@ void Assembler::assessAlignments(
     markerCountHistogram.writeToHtml(html, histogramSize);
     html << "<br><strong>Aligned Fraction Distribution</strong>";
     alignedFractionHistogram.writeToHtml(html, histogramSize);
-    html << "<br><strong>Alignment Quantity Distribution</strong>";
+    html << "<br><strong>Number of Alignments Found per Read</strong>";
     nAlignmentsHistogram.writeToHtml(html, histogramSize);
     html << "<br><strong>Ratio of stored to found alignments</strong>";
     html << "<br>" << double(allStoredAlignments.size())/double(allAlignments.size());
--- a/src/AssemblerHttpServer-Alignments.cpp
+++ b/src/AssemblerHttpServer-Alignments.cpp
@@ -1272,8 +1272,9 @@ void Assembler::assessAlignments(
     getParameterValue(request, "bandExtend", computeAllAlignmentsData.bandExtend);
 
     html << "<h1>Alignment statistics</h1>";
-    html << "<p>This page enables sampling from the pool of reads and computing alignments for each read in the"
-            "sample. Once alignment finishes, stats can be generated and used to evaluate Shasta parameters."
+    html << "<p>This page enables sampling from the pool of reads and computing alignments for each read in the sample "
+            "against all other reads in this assembly. This can be slow. Once alignment finishes, stats can be "
+            "generated and used to evaluate Shasta parameters."
             "<br>";
 
     // Write the form.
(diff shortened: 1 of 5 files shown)


rlorigro:
last push was a rebase

bagashe:
This is very cool. I will do a detailed review later tonight. 

Can we compute this histogram during a regular Shasta run and dump it in the ShastaRun directory? I could use it in the feedback script for automated feedback.

rlorigro:
Yeah you could totally produce something like this from the stored alignments. The difference will be that the stored alignments are only calculated from the subset of reads that were paired by the LowHash algorithm, and the distributions will have a hard cutoff at `alignedFraction = 0.4` and `markerCount = 200` or whatever your configuration is set to.

We can figure out the details in another PR

src/AssemblerHttpServer-Alignments.cpp
```diff
@@ -1182,6 +1183,255 @@ void Assembler::computeAllAlignments(
 }
 
 
+void Assembler::sampleReads(vector<OrientedReadId>& sample, uint64_t n){
+    sample.clear();
+
+    while (sample.size() < n) {
+        // Randomly select a read in the read set
+        uint16_t readId = uint16_t(rand() % reads.size());
```
bagashe, line 1191: A `uint16_t` can hold a max value of 65536. So this sampling will always choose from the first 65536 reads. Is that intentional? `OrientedReadId` takes a `uint32_t` readId.

src/AssemblerHttpServer-Alignments.cpp
```diff
@@ -1182,6 +1183,255 @@ void Assembler::computeAllAlignments(
...
+
+        // Randomly select an orientation
+        uint16_t strand = uint16_t(rand() % 2);
+
+        const OrientedReadId r(readId, strand);
+
+        sample.push_back(r);
```
bagashe, line 1198: [Nit-pick] 
```
const OrientedReadId r(readId, strand);
sample.push_back(r);
```
creates an object and copies it into the vector. I would assume that the compiler is smart enough to optimize this away. 

You could also do the following to use C++ move semantics explicitly (which will move the object into the vector instead of copying it).

```
sample.push_back(OrientedReadId(readId, strand));
```

Feel free to ignore this comment if you like your style. I am 90% certain the compiler will do the right thing here.

src/AssemblerHttpServer-Alignments.cpp
```diff
@@ -1182,6 +1183,255 @@ void Assembler::computeAllAlignments(
...
+    }
+}
+
+
+void Assembler::sampleReads(vector<OrientedReadId>& sample, uint64_t n, uint64_t minLength, uint64_t maxLength){
+    sample.clear();
+
+    while (sample.size() < n) {
+        // Randomly select a read in the read set
+        uint16_t readId = uint16_t(rand() % reads.size());
```
bagashe, line 1208: Same here.

src/AssemblerHttpServer-Alignments.cpp
```diff
@@ -1182,6 +1183,255 @@ void Assembler::computeAllAlignments(
...
+
+        // Only update the sample of reads if this read passes the length criteria
+        if(length >= minLength and length <= maxLength) {
+            sample.push_back(r);
+        }
+    }
+}
+
+
+// Compute alignments on an oriented read against
```
bagashe, line 1230: For a full human genome, aligning one read with all others takes 40+ seconds on a 128 thread machine. Won't this take minutes for a small sample size of 100?  

If we can't wait that long in a browser, may be we could have this function spawn a thread that generates this data and writes it out to the ShastaRun directory. Even better, I would do this during the main Shasta run and dump the histograms in files for later analysis. Adding 5 minutes to the assembly time seems totally worth it, if the eventual goal is to pick these thresholds automatically anyways. 

What do you think? If you agree, I can help you move this code to the main assembly process.

src/AssemblerHttpServer-Alignments.cpp
rlorigro: Hmm definitely a mistake, i thought my compiler was giving me conversion warnings with uint32_t but I must have misread it.

src/AssemblerHttpServer-Alignments.cpp
rlorigro: For this one I am aware that it would copy without optimization, but as you said there is no point in using `emplace_back`, `mov`, or a one-liner to try to manually optimize it because the compiler will do it.

paoloczi: The compiler will do the right optimization here, and both ways are clear and readable, although I personally prefer @bagashe's way  because it does not introduce an extra named variable. In any event most of the cost by far will go in computing the alignments, so the performance of these preparation steps does not really matter.

src/AssemblerHttpServer-Alignments.cpp
paoloczi: Please use `ReadId`, currently defined as `uint32_t`. This is mostly for better readability, although the theoretical possibility that this will change one day in the distant future exists. Also I prefer to define as `const` local variables that are filled in on assignment and never changed.

src/AssemblerHttpServer-Alignments.cpp
paoloczi: On a large run and depending on alignment parameters, read length distribution, and machine size, this can take several minutes per read. So if we wanted to do it for 50 reads it could take hours. This would be too much time to add to the assembly by default, but we could still do it in parallel for a small number of reads if we can squeeze it in one of our entirely sequential phases. Even in this case, I would still do this optionally and not by default, at least for now.

rlorigro: Another alternative would be to trust the LowHash step, and do this thresholding on the reads that have been paired already, instead of 1-to-all.

paoloczi: A couple of comments on the histograms. Each bin is labeled with one value (e. g. 0.55) and it is not clear what the bin interval is.Does 0.55 mean "0.55 to 0.60", "0.50 to 0.55", or "0.525 to 0.575"? We have to make this clear, either by adding a second column to characterize each bin (and labeling the two colummns "Min" and "Max" or something similar), or with a comment (I prefer the former solution).

The title of the second histogram is confusing. What does "Aligned Quantity distribution" mean? I imagine it is the number of alignments found for each read. This needs to be clarified.

Don't we also need a histogram for the number of aligned markers?

rlorigro: The histogram for number of aligned markers is included, but not shown in the picture because it is long (0-3000). 

Aligned Quantity distribution is the number of alignments found for each read. I agree the name is not clear.

Ok, the binning is easy to do

paoloczi:
A couple more comments on the page.

* It should have a title (`<h1>`). Something like "Alignment statistics"? And perhaps a blurb explaining what it does.
*  What is the meaning of the minimum and maximum read lengths? There are three possibilities: bases of raw sequence (the original read), bases of RLE sequence, or markers.

src/AssemblerHttpServer-Alignments.cpp
```diff
@@ -1182,6 +1183,255 @@ void Assembler::computeAllAlignments(
...
+
+void Assembler::sampleReads(vector<OrientedReadId>& sample, uint64_t n, uint64_t minLength, uint64_t maxLength){
+    sample.clear();
+
+    while (sample.size() < n) {
+        // Randomly select a read in the read set
+        uint16_t readId = uint16_t(rand() % reads.size());
+
+        // Randomly select an orientation
+        uint16_t strand = uint16_t(rand() % 2);
```
paoloczi, line 1211: There is also a `Strand` type. And I would make this `const`, and do the same for local variables initialized once and not changed. It actually helps avoid bugs.

src/AssemblerHttpServer-Alignments.cpp
```diff
@@ -1182,6 +1183,255 @@ void Assembler::computeAllAlignments(
...
+        // Vectors to contain markers sorted by kmerId.
+        vector<MarkerWithOrdinal> markers0SortedByKmerId;
+        vector<MarkerWithOrdinal> markers1SortedByKmerId;
+        getMarkersSortedByKmerId(orientedReadId, markers0SortedByKmerId);
+
+        // Compute the alignments in parallel.
+        computeAllAlignmentsData.orientedReadId0 = orientedReadId;
+        const size_t threadCount = std::thread::hardware_concurrency();
+        computeAllAlignmentsData.threadAlignments.resize(threadCount);
+        const size_t batchSize = 1000;
```
paoloczi, line 1357: With this batch size, the first thread grabs all the reads and the remaining threads do nothing. For this application, the batch size should be 1.

src/AssemblerHttpServer-Alignments.cpp
```diff
@@ -1182,6 +1183,255 @@ void Assembler::computeAllAlignments(
...
+                    computeAllAlignmentsData.threadAlignments[threadId];
+            copy(threadAlignments.begin(), threadAlignments.end(), back_inserter(alignments));
+        }
+        computeAllAlignmentsData.threadAlignments.clear();
+        sort(alignments.begin(), alignments.end(),
+             OrderPairsByFirstOnly<OrientedReadId, AlignmentInfo>());
+
+        // Loop over the STORED alignments that this oriented read is involved in, with the proper orientation.
+        const vector< pair<OrientedReadId, AlignmentInfo> > storedAlignments =
+                findOrientedAlignments(orientedReadId);
```
paoloczi, line 1378: If you only need the number of alignments, you can use `alignmentTable.size(orientedReadId.getValue())` instead.

rlorigro: Yea, I also put them all in a vector because for the next step I want to do a table of comparisons for stored vs found alignments. I think the comparison table will help with dead ends (e.g. # of overhangs). Waiting for the next PR if that's alright

src/Histogram.hpp
```diff
@@ -43,4 +48,34 @@ class shasta::Histogram : public vector<size_t> {
 };
 
 
+// A histogram class that calculates bin size and finds the appropriate bin to increment given an x value
```
paoloczi, line 51: Would it make sense to use the [Boost Accumulators library](https://www.boost.org/doc/libs/1_73_0/doc/html/accumulators.html) instead? It has a [`density`](https://www.boost.org/doc/libs/1_73_0/doc/html/accumulators/user_s_guide.html#accumulators.user_s_guide.the_statistical_accumulators_library.density) feature that computes a histogram.

src/AssemblerHttpServer-Alignments.cpp
paoloczi: > Another alternative would be to trust the LowHash step, and do this thresholding on the reads that have been paired already, instead of 1-to-all.

I don't see much point in this, because one of the main purposes of doing this is to check that the LowHash step is doing its job.

src/Histogram.hpp
rlorigro: ah I didnt know about that. It looks like they have tons of other incremental stats objects

src/AssemblerHttpServer-Alignments.cpp
paoloczi: > The histogram for number of aligned markers is included, but not shown in the picture because it is long (0-3000).

You can bin it, like the other hsistograms.

rlorigro: > I don't see much point in this, because one of the main purposes of doing this is to check that the LowHash step is doing its job.

It is binned, but it is still long because the interval size needs to be small enough to see the distribution clearly on the lower end.

> I don't see much point in this, because one of the main purposes of doing this is to check that the LowHash step is doing its job.

Hmm I was thinking that the purpose was to remove the need to specify Align.minMarkerCount and Align.minAlignedFraction as parameters in the config. I think it would be fair to estimate the distribution of the alignments given the candidate pairs for this purpose.

rlorigro: example:
image

paoloczi: > Hmm I was thinking that the purpose was to remove the need to specify Align.minMarkerCount and Align.minAlignedFraction as parameters in the config. I think it would be fair to estimate the distribution of the alignments given the candidate pairs for this purpose.

This is a good point. There is a scenario where relying on the stored alignments would be useful.

paoloczi: > It is binned, but it is still long because the interval size needs to be small enough to see the distribution clearly on the lower end.

So perhaps use a non-uniform binning. Perhaps uniform in log space?

rlorigro:
Reverted boost accumulator edits because it did not have the option for specifying min/max, and it appeared to be failing for  decimals between 0 and 1

src/AssemblerHttpServer-Alignments.cpp
```diff
@@ -1182,6 +1183,262 @@ void Assembler::computeAllAlignments(
...
+    computeAllAlignmentsData.gapScore = httpServerData.assemblerOptions->alignOptions.gapScore;
+    getParameterValue(request, "gapScore", computeAllAlignmentsData.gapScore);
+    computeAllAlignmentsData.downsamplingFactor = httpServerData.assemblerOptions->alignOptions.downsamplingFactor;
+    getParameterValue(request, "downsamplingFactor", computeAllAlignmentsData.downsamplingFactor);
+    computeAllAlignmentsData.bandExtend = httpServerData.assemblerOptions->alignOptions.bandExtend;
+    getParameterValue(request, "bandExtend", computeAllAlignmentsData.bandExtend);
+
+    html << "<h1>Alignment statistics</h1>";
+    html << "<p>This page enables sampling from the pool of reads and computing alignments for each read in the"
+            "sample. Once alignment finishes, stats can be generated and used to evaluate Shasta parameters."
```
paoloczi, line 1276: For clarity I would change the message to "...for each read in the sample against all other reads in this assembly. This can be slow."

src/AssemblerHttpServer-Alignments.cpp
```diff
@@ -1182,6 +1183,262 @@ void Assembler::computeAllAlignments(
...
+    IterativeHistogram nAlignmentsHistogram(0,200,20);
+
+    vector<pair<OrientedReadId, AlignmentInfo> > allAlignments;
+    vector<pair<OrientedReadId, AlignmentInfo> > allStoredAlignments;
+
+    for (auto& orientedReadId: sampledReads) {
+        // Vectors to contain markers sorted by kmerId.
+        vector<MarkerWithOrdinal> markers0SortedByKmerId;
+        vector<MarkerWithOrdinal> markers1SortedByKmerId;
+        getMarkersSortedByKmerId(orientedReadId, markers0SortedByKmerId);
```
paoloczi, line 1353: This is only necessary for alignment method 0.

src/Histogram.hpp
```diff
@@ -43,4 +48,34 @@ class shasta::Histogram : public vector<size_t> {
...
+class shasta::IterativeHistogram{
```
paoloczi, line 52: Why is this called an `IterativeHistogram`? I see nothing iterative here.

src/Histogram.hpp
```diff
@@ -43,4 +48,34 @@ class shasta::Histogram : public vector<size_t> {
...
+    const double binSize;
+    vector<uint64_t> histogram;
+    bool unboundedLeft;
+    bool unboundedRight;
+
+    /// Methods ///
+    IterativeHistogram(
+            double start,
+            double stop,
+            size_t nBins,
```
paoloczi, line 67: Calling this `binCount` instead of `nBins` would be more consistent with typical naming in Shasta code.

src/Histogram.hpp
paoloczi: Perhaps just call it `Histogram2`, and at some point we can remove the more basic, existing `Histogram` class using `Histogram2` instead, then finally rename `Histogram2` to `Histogram`.

rlorigro: Maybe "incremental" or "accumulative" would have made more sense, but Histogram2 is fine

src/AssemblerHttpServer-Alignments.cpp
rlorigro: The page allows the user to pick any method

paoloczi: Is it possible to skip the calls to `getMarkersSortedByKmerId` if the alignment method is not zero?

src/Histogram.hpp
paoloczi: Yes, and if we do what I said above, it will eventually be renamed `Histogram` once the current `Histogram` class goes away.

src/AssemblerHttpServer-Alignments.cpp
rlorigro: Sure. This is all copied over from the standard alignments page, including the batchSize comment you made earlier. Should I update both pages?

PR #171, opened 2020-07-24
Script to generate starter config and script to provide feedback.

bagashe:
Version 1 of two helper scripts. The goal with these isn't to be perfect. It's to help users do some basic experimentation themselves by nudging them in the right direction.


scripts/GenerateConfig.py
```diff
@@ -0,0 +1,229 @@
...
+                    yield line
+                
+
+
+def getMinReadLength(inputFiles, genomeSize, baseCallerId):
+    # Reads map to a bin that is int(read-length / 1000).
+    binnedHistogramOfBases = [0] * 1000
+    startTime = time.time()
+
+    for inputFile in inputFiles:
```
bagashe, line 61: Instead of reading the entire file, I can read a smaller fraction of it (say 10%) and compute my stats based on that. I plan to do that in a separate PR shortly.

bagashe:
This PR is on hold, while I transfer some logic to core Shasta.

bagashe:
Removed the things that have been moved into the Shasta executable itself. I have add the necessary disclaimers in the `usage` text. I think this is good enough for the first version.

PR #174, opened 2020-07-31
[Refactor] Reads class

bagashe:
1. `reads`, `readRepeatCounts`, `readNames`, `readMetaData` & `readFlags` are now wrapped in a class named `Reads`.
2. Moved the Read length histogram computation to the `Reads` class as well. This will allow us to do something based on desired coverage (in a later PR).
3. Consolidated creation of `readFlags` with the others. `ReadLoader` now prepares (resizes) `readFlags` after it is done populating `reads` & `readRepeatCounts`. 

In follow-up PR(s)
1. Make `reads`, `readRepeatCounts`, `readNames`, `readMetaData` & `readFlags` private. This would require creating a const version of `LongBaseSequenceView` (and potentially other things).
2. Switch the data members to pointers so that they can be swapped after implementing support for `--Reads.desiredCoverage`

### Test Plan
Ran E-coli assembly before and after the change. `diff `d the two AssemblySummary.json files to verify that there was no material change in the assembly.

Explored the assembly using the HTTP server to ensure those code paths work as well.

### DIFF
--- a/src/Assembler.hpp
+++ b/src/Assembler.hpp
@@ -23,6 +23,7 @@
 #include "ReadGraph.hpp"
 #include "ReadFlags.hpp"
 #include "ReadId.hpp"
+#include "Reads.hpp"
 
 // Standard library.
 #include "memory.hpp"
@@ -47,6 +48,7 @@ namespace shasta {
     class LocalAlignmentGraph;
     class LocalReadGraph;
     class SegmentGraph;
+    class Reads;
 
 #ifdef SHASTA_HTTP_SERVER
     class LocalMarkerGraph;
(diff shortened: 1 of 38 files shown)


src/Assembler.hpp
```diff
@@ -47,6 +48,7 @@ namespace shasta {
     class LocalAlignmentGraph;
     class LocalReadGraph;
     class SegmentGraph;
+    class Reads;
```
paoloczi, line 51: This forward definition should not be necessary, if we are including `Reads.hpp`.

src/AssemblerAlign.cpp
```diff
@@ -733,10 +734,13 @@ void Assembler::flagPalindromicReadsThreadFunction(size_t threadId)
 
     // Loop over all batches assigned to this thread.
     uint64_t begin, end;
+    reads.assertReadsAndFlagsOfSameSize();
+    ReadId readCount = reads.readCount();
```
paoloczi, line 738: `const`?

src/AssemblerConflictReadGraph.cpp
```diff
@@ -46,7 +46,7 @@ void Assembler::createConflictReadGraph(
 
     // Initialize the conflict read graph.
     conflictReadGraph.createNew(largeDataName("ConflictReadGraph"), largeDataPageSize);
-    conflictReadGraph.createVertices(readCount());
+    conflictReadGraph.createVertices(reads.readCount());
```
paoloczi, line 49: Perhaps it would have made sense to keep `Assembler::readCount()` and let it simply forward to `class Reads`. Then code like this would not have been affected.

paoloczi: But I am not proposing that we make this change now.

Issue #172, opened 2020-07-28
Shasta integration with snakemake


BCArg:
I have written a denovo assembly pipeline using snakemake where the first step is to build a draft assembly using shasta. Though I am wondering whether Shasta can be integrated into a snakemake pipeline at all.

The error that I am getting is that the output directory already existed prior to running shasta (I am getting the same error for all four samples I wanted to call the snakemake on)

```
2020-Jul-28 09:28:40.641060 Terminated after catching a runtime error exception:
Assembly directory BORDref248_barcode10/ShastaRun/ already exists.
Remove it or use --assemblyDirectory to specify a different assembly directory.
```

Those directories did not exist prior to calling snakemake, though they were specified in the the
1.  `output` directive of the `shasta` rule in the snakemake (please see below)
2. `--assemblyDirectory` flag in the `shell` directive of the shasta rule

as far as I am aware, the output directories as specified in `assemblyDirectory` were created by snakemake prior to executing the commands under `shell`. So given that shasta throws an error when the output directory already exists, I am now wondering if it can be integrated into a denovo assembly pipeline in snakemake at all.

Below  is the shasta rule:

```
rule shasta:
    input:
        fqdir + "{sample}.fastq"
    output:
        "{sample}/ShastaRun/Assembly.fasta",
        "{sample}/ShastaRun/Assembly.gfa",
        "{sample}/ShastaRun/Assembly-BothStrands.gfa",
        "{sample}/ShastaRun/shasta.conf",
        "{sample}/ShastaRun/ReadLengthHistogram.csv",
        "{sample}/ShastaRun/Binned-ReadLengthHistogram.csv",
        "{sample}/ShastaRun/PalindromicReads.csv",
        "{sample}/ShastaRun/ReadLowHashStatistics.csv",
        "{sample}/ShastaRun/LowHashBucketHistogram.csv",
        "{sample}/ShastaRun/ReadGraphComponents.csv",
        "{sample}/ShastaRun/MarkerGraphVertexCoverageHistogram.csv",
        "{sample}/ShastaRun/MarkerGraphEdgeCoverageHistogram.csv",
        "{sample}/ShastaRun/AssemblyGraphChainLengthHistogram.csv",
        "{sample}/ShastaRun/AssemblyGraph-Final.dot",
        "{sample}/ShastaRun/AssemblySummary.csv",
        "{sample}/ShastaRun/ReadSummary.csv",
        "{sample}/ShastaRun/index.html",
[...]

PR #175, opened 2020-08-03
Allow an empty assembly directory to exist before the assembly starts.

bagashe:
This will allow integration with tools like snakemake 

https://github.com/chanzuckerberg/shasta/issues/172

###  Test Plan
1. Ran an assembly where --assemblyDirectory does not exist (No error)
2. Ran an assembly where --assemblyDirectory exists and is not empty (Throws an error)
3. Ran an assembly where --assemblyDirectory exists and is empty (No error)

### DIFF
--- a/srcMain/main.cpp
+++ b/srcMain/main.cpp
@@ -229,8 +229,16 @@ void shasta::main::assemble(
 
 
     // Create the run output directory. If it exists and is not empty then stop.
-    bool dirExists = filesystem::isDirectory(assemblerOptions.commandLineOnlyOptions.assemblyDirectory);
-    if (dirExists) {
+    bool exists = filesystem::exists(assemblerOptions.commandLineOnlyOptions.assemblyDirectory);
+    bool isDir = filesystem::isDirectory(assemblerOptions.commandLineOnlyOptions.assemblyDirectory);
+    if (exists) {
+        if (!isDir) {
+            throw runtime_error(
+                assemblerOptions.commandLineOnlyOptions.assemblyDirectory +
+                " already exists and is not a directory.\n"
+                "Use --assemblyDirectory to specify a different assembly directory."
+            );
+        }
         bool isEmpty = filesystem::directoryContents(assemblerOptions.commandLineOnlyOptions.assemblyDirectory).empty();
         if (!isEmpty) {
             throw runtime_error(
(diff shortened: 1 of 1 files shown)


srcMain/main.cpp
```diff
@@ -228,16 +228,20 @@ void shasta::main::assemble(
...
 
-    // Create the run the output directory. If it exists, stop.
-    if(filesystem::exists(assemblerOptions.commandLineOnlyOptions.assemblyDirectory)) {
-        throw runtime_error(
-            "Assembly directory " +
-            assemblerOptions.commandLineOnlyOptions.assemblyDirectory +
-            " already exists.\n"
-            "Remove it or use --assemblyDirectory to specify a different assembly directory.");
+    // Create the run the output directory. If it exists and is not empty then stop.
+    bool dirExists = filesystem::exists(assemblerOptions.commandLineOnlyOptions.assemblyDirectory);
```
paoloczi, line 232: Also check that it is a directory, with a call to `shasta::filesystem::isDirectory`?

bagashe: Looks like `shasta::filesystem::isDirectory` does the exists check as well. Great.

srcMain/main.cpp
```diff
@@ -228,16 +228,20 @@ void shasta::main::assemble(
...
 
-    // Create the run the output directory. If it exists, stop.
-    if(filesystem::exists(assemblerOptions.commandLineOnlyOptions.assemblyDirectory)) {
-        throw runtime_error(
-            "Assembly directory " +
-            assemblerOptions.commandLineOnlyOptions.assemblyDirectory +
-            " already exists.\n"
-            "Remove it or use --assemblyDirectory to specify a different assembly directory.");
+    // Create the run output directory. If it exists and is not empty then stop.
+    bool dirExists = filesystem::isDirectory(assemblerOptions.commandLineOnlyOptions.assemblyDirectory);
```
paoloczi, line 232: If the specified `assemblyDirectory` exists but is a file rather than a directory, `isDirectory` returns `false` and the code below will attempt to create a directory with that same name, which will fail because of existing file with the same name. Perhaps it is best to check for that explicitly.

bagashe: Cool

PR #176, opened 2020-08-04
Make data members of the Reads class, private.

bagashe:
`LongBaseSequences::operator[]` needs a `const` overload to cleanup dependent code. There're a couple different ways to accomplish that. This one feels the cleanest - I am open to be convinced otherwise. 

While tightening the `const` screws, I had to do the modifications in `span.hpp` as well. 

### Test Plan
1. Ran an E-Coli assembly.
2. Explored said assembly using the HTTP server.

### DIFF
--- a/src/LongBaseSequence.hpp
+++ b/src/LongBaseSequence.hpp
@@ -43,15 +43,22 @@ class shasta::LongBaseSequenceView {
 public:
     uint64_t* begin;
     uint64_t baseCount;
+    bool readOnly;
 
     LongBaseSequenceView(
         uint64_t* begin,
         uint64_t baseCount) :
-        begin(begin), baseCount(baseCount)
+        begin(begin), baseCount(baseCount), readOnly(false)
     {}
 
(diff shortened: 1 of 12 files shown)


src/LongBaseSequence.hpp
```diff
@@ -43,15 +43,22 @@ class shasta::LongBaseSequenceView {
...
+
+    LongBaseSequenceView(
+        const uint64_t* begin,
+        const uint64_t baseCount) :
+        begin(const_cast<uint64_t*>(begin)), baseCount(baseCount), readOnly(true)
     {}
 
     LongBaseSequenceView() :
-        begin(0), baseCount(0)
+        begin(0), baseCount(0), readOnly(false)
```
paoloczi, line 61: Perhaps set `readOnly(true)` here, to make it more likely that an assertion occurs if attempting to use a default-constructed `LongBaseSequenceView`.

bagashe: Good point. Will try it out.

PR #177, opened 2020-08-06
Reduce coverage to the desired value if applicable.

bagashe:
Desired coverage, as a number of raw bases, can be set using the new configuration parameter -  `Reads.desiredCoverage`. This value will be auto-generated by the `GenerateConfig.py` script (different PR). `Reads.minReadLength` should now default to a potentially smaller value (from the current default value of 10,000). 

- If there isn't enough coverage available with reads longer than `Reads.minReadLength`, the program will abort. 
- If more than desired coverage is available with reads longer than `Reads.minReadLength`, then Shasta will increase the value of `minReadLength` to arrive close to the desired coverage

### DIFF
--- a/src/AssemblerOptions.cpp
+++ b/src/AssemblerOptions.cpp
@@ -201,6 +201,11 @@ void AssemblerOptions::addConfigurableOptions()
         default_value(10000),
         "Read length cutoff. Shorter reads are discarded.")
 
+        ("Reads.desiredCoverage",
+        value<uint64_t>(&readsOptions.desiredCoverage)->
+        default_value(0),
+        "Desired coverage as a number of bases. Coverage will be reduced to this if necessary.")
+
         ("Reads.noCache",
         bool_switch(&readsOptions.noCache)->
         default_value(false),
@@ -594,6 +599,7 @@ void AssemblerOptions::ReadsOptions::write(ostream& s) const
 {
     s << "[Reads]\n";
     s << "minReadLength = " << minReadLength << "\n";
+    s << "desiredCoverage = " << desiredCoverage << "\n";
     s << "noCache = " <<
         convertBoolToPythonString(noCache) << "\n";
     palindromicReads.write(s);
--- a/srcMain/main.cpp
+++ b/srcMain/main.cpp
@@ -478,6 +478,7 @@ void shasta::main::assemble(
         assembler.addReads(
             inputFileName,
             assemblerOptions.readsOptions.minReadLength,
+            assemblerOptions.readsOptions.desiredCoverage,
             assemblerOptions.readsOptions.noCache,
             threadCount);
     }
@@ -488,13 +489,10 @@ void shasta::main::assemble(
     cout << timestamp << "Done loading reads from " << inputFileNames.size() << " files." << endl;
     cout << "Read loading took " << seconds(t1-t0) << "s." << endl;
 
-
-    // Create a histogram of read lengths.
+    // Log the read length histogram.
     assembler.histogramReadLength("ReadLengthHistogram.csv");
 
-
-
-    // Select the k-mers that will be used as markers.
+        // Select the k-mers that will be used as markers.
     switch(assemblerOptions.kmersOptions.generationMethod) {
     case 0:
         assembler.randomlySelectKmers(
--- a/src/Assembler.hpp
+++ b/src/Assembler.hpp
@@ -171,6 +171,7 @@ public:
     void addReads(
         const string& fileName,
         uint64_t minReadLength,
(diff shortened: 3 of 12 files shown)


paoloczi:
Your initial comment on this PR says:

_`Reads.minReadLength` should now default to a potentially smaller value (from the current default value of 10,000)._

Why would this be the case? Can you clarify?

srcMain/main.cpp
```diff
@@ -488,13 +489,10 @@ void shasta::main::assemble(
     cout << timestamp << "Done loading reads from " << inputFileNames.size() << " files." << endl;
     cout << "Read loading took " << seconds(t1-t0) << "s." << endl;
 
-
-    // Create a histogram of read lengths.
+    // Log the read length histogram.
```
paoloczi, line 492: I thought "Log" means "write to stdout". So this comment seems a bit misleading.

src/AssemblerOptions.cpp
```diff
@@ -201,6 +201,11 @@ void AssemblerOptions::addConfigurableOptions()
         default_value(10000),
         "Read length cutoff. Shorter reads are discarded.")
 
+        ("Reads.desiredCoverage",
+        value<uint64_t>(&readsOptions.desiredCoverage)->
+        default_value(0),
+        "Desired coverage as a number of bases. Coverage will be reduced to this if necessary.")
```
paoloczi, line 207: After the PR is merged, I will update the documentation and `shasta.conf` to reflect this new option. I will also expand this help message a bit, consistently with the documentation changes, to explain the meaning of specifying zero coverage.

bagashe:
> Your initial comment on this PR says:
> 
> _`Reads.minReadLength` should now default to a potentially smaller value (from the current default value of 10,000)._
> 
> Why would this be the case? Can you clarify?

If desired coverage is specified, then `minReadLength` can be conceptually repurposed to specify the lower bound of what an acceptable read length is for Shasta, in general. For example, if Shasta's algorithm works best when read lengths are longer than 3000 bases, then that's what we can set it to. In other words, it can now be a function of something other than available coverage.

srcMain/main.cpp
bagashe: Does it? In my mind, it means, log it to the logging system, which can be anything really. I can clarify as it is confusing to you.

src/AssemblerOptions.cpp
bagashe: Good point. The assumption is that this flag will be set by the `GenerateConfig.py` script, which asks the user for genome size in mega bases. I do not anticipate users entering this value manually.

srcMain/main.cpp
paoloczi: True, but the only logging system we have is writing to `stdout`. Although you could consider our various `csv` output files a form of logging. Minor point anyway - your choice.

paoloczi:
_"if Shasta's algorithm works best when read lengths are longer than 3000 bases, then that's what we can set it to."_

Oh I see. In practice, at least for human genomes, 10 Kb seems a good value - probably due to the existence of LINE repeats in the human genome which are 6 Kb long. In other words, if you only have coverage 30X in reads longer than 10 Kb, it is probably not a good idea to increase coverage by allowing shorter reads. What you really need is more coverage.

The more common usage pattern, at least for human genomes, will be the case where you have a lot of coverage - say 200X - and you want to reduce it to something like 70X by **increasing** `minReadLength`.

src/AssemblerOptions.cpp
paoloczi: There will certainly be cases where users don't want to use `GenerateConfig.py` for whatever reason, but want to be able to specify desired coverage manually (in bases, without reference to a genome length). This would be a very useful option that I have missed many times when running Shasta assemblies. Requiring all the zeros would make it less usable but still useful.

bagashe: Makes sense. I will support suffixes in the other PR I put up about this change (Where we swap the entire Reads object in the Assembler). Should we support anything other than `G/Gbp` and `M/Mbp`?

paoloczi: I would also do `K` - it could be useful for very small viral genomes. Although in that case that option is not as needed as for the longer genomes, because there would not be too many zeros. 

If you allow `Gbp/Mbp/Kbp`, I would also allow `Gb/Mb/Kb`. 

Those prefixes should be interpreted as powers of 10, not powers of 1024. I will clarify that in the documentation too.

bagashe:
Closing this in favor of https://github.com/chanzuckerberg/shasta/pull/178.

PR #178, opened 2020-08-10
Reduce coverage to desired value if applicable.

bagashe:
Desired coverage, as a number of bases, can be set using the new configuration parameter - Reads.desiredCoverage. This value will be auto-generated by the GenerateConfig.py script (different PR). 

- If there isn't enough coverage available with reads longer than Reads.minReadLength, the program will abort.
- If more than desired coverage is available with reads longer than Reads.minReadLength, then Shasta will increase the value of minReadLength to arrive close to the desired coverage

`Reads.desiredCoverage` can take values specified as 
- raw number of bases. E.g `--Reads.desiredCoverage 2000000000`
- kilo bases (K, Kb, Kbp). E.g. `--Reads.desiredCoverage 2000000Kb`
- mega bases (M, Mb, Mbp). E.g. `--Reads.desiredCoverage 2000Mb`
- giga bases (G, Gb, Gbp). E.g. `--Reads.desiredCoverage 2Gb`

**Performance hit**: For `HG002-Guppy-3.6.0.fasta` dropping coverage to `120G` adds a little over 2 minutes to the assembly time.

### DIFF
--- a/src/AssemblerAlign.cpp
+++ b/src/AssemblerAlign.cpp
@@ -38,8 +38,8 @@ void Assembler::alignOrientedReads(
     uint32_t maxMarkerFrequency
 )
 {
-    reads.checkReadsAreOpen();
-    reads.checkReadNamesAreOpen();
+    reads->checkReadsAreOpen();
+    reads->checkReadNamesAreOpen();
     checkMarkersAreOpen();
 
 
@@ -62,9 +62,9 @@ void Assembler::alignOrientedReads(
     uint32_t leftTrim;
     uint32_t rightTrim;
     tie(leftTrim, rightTrim) = alignmentInfo.computeTrim();
-    cout << orientedReadId0 << " has " << reads.getRead(orientedReadId0.getReadId()).baseCount;
+    cout << orientedReadId0 << " has " << reads->getRead(orientedReadId0.getReadId()).baseCount;
     cout << " bases and " << markersSortedByKmerId[0].size() << " markers." << endl;
-    cout << orientedReadId1 << " has " << reads.getRead(orientedReadId1.getReadId()).baseCount;
+    cout << orientedReadId1 << " has " << reads->getRead(orientedReadId1.getReadId()).baseCount;
     cout << " bases and " << markersSortedByKmerId[1].size() << " markers." << endl;
     cout << "The alignment has " << alignmentInfo.markerCount;
     cout << " markers. Left trim " << leftTrim;
@@ -146,7 +146,7 @@ void Assembler::alignOverlappingOrientedReads(
 )
 {
     // Check that we have what we need.
-    reads.checkReadsAreOpen();
+    reads->checkReadsAreOpen();
     checkMarkersAreOpen();
     checkAlignmentCandidatesAreOpen();
 
@@ -254,7 +254,7 @@ void Assembler::computeAlignments(
     cout << alignmentCandidates.candidates.size() << " alignment candidates." << endl;
 
     // Check that we have what we need.
-    reads.checkReadsAreOpen();
+    reads->checkReadsAreOpen();
     checkKmersAreOpen();
     checkMarkersAreOpen();
     checkAlignmentCandidatesAreOpen();
@@ -527,7 +527,7 @@ void Assembler::accessCompressedAlignments()
 void Assembler::computeAlignmentTable()
 {
     alignmentTable.createNew(largeDataName("AlignmentTable"), largeDataPageSize);
-    alignmentTable.beginPass1(ReadId(2 * reads.readCount()));
+    alignmentTable.beginPass1(ReadId(2 * reads->readCount()));
     for(const AlignmentData& ad: alignmentData) {
         const auto& readIds = ad.readIds;
         OrientedReadId orientedReadId0(readIds[0], 0);
@@ -558,7 +558,7 @@ void Assembler::computeAlignmentTable()
 
     // Sort each section of the alignment table by OrientedReadId.
     vector< pair<OrientedReadId, uint32_t> > v;
-    for(ReadId readId0=0; readId0<reads.readCount(); readId0++) {
+    for(ReadId readId0=0; readId0<reads->readCount(); readId0++) {
         for(Strand strand0=0; strand0<2; strand0++) {
             const OrientedReadId orientedReadId0(readId0, strand0);
 
@@ -679,10 +679,10 @@ void Assembler::flagPalindromicReads(
     flagPalindromicReadsData.deltaThreshold = deltaThreshold;
 
     // Reset all palindromic flags.
-    reads.assertReadsAndFlagsOfSameSize();
-    const ReadId readCount = reads.readCount();
+    reads->assertReadsAndFlagsOfSameSize();
+    const ReadId readCount = reads->readCount();
     for(ReadId readId=0; readId<readCount; readId++) {
-        reads.setPalindromicFlag(readId, false);
+        reads->setPalindromicFlag(readId, false);
     }
 
     // Do it in parallel.
@@ -692,7 +692,7 @@ void Assembler::flagPalindromicReads(
     // Count the reads flagged as palindromic.
     size_t palindromicReadCount = 0;
     for(ReadId readId=0; readId<readCount; readId++) {
-        if(reads.getFlags(readId).isPalindromic) {
+        if(reads->getFlags(readId).isPalindromic) {
             ++palindromicReadCount;
         }
     }
@@ -708,7 +708,7 @@ void Assembler::flagPalindromicReads(
     // palindromic reads is around 1e-4.
     ofstream csvOut("PalindromicReads.csv");
     for(ReadId readId=0; readId<readCount; readId++) {
-        if(reads.getFlags(readId).isPalindromic) {
+        if(reads->getFlags(readId).isPalindromic) {
             csvOut << readId << "\n";
         }
     }
@@ -737,8 +737,8 @@ void Assembler::flagPalindromicReadsThreadFunction(size_t threadId)
 
     // Loop over all batches assigned to this thread.
     uint64_t begin, end;
-    reads.assertReadsAndFlagsOfSameSize();
-    ReadId readCount = reads.readCount();
+    reads->assertReadsAndFlagsOfSameSize();
+    ReadId readCount = reads->readCount();
 
     while(getNextBatch(begin, end)) {
         if((begin%1000000) == 0) {
@@ -783,7 +783,7 @@ void Assembler::flagPalindromicReadsThreadFunction(size_t threadId)
             }
 
             // If we got here, mark the read as palindromic.
-            reads.setPalindromicFlag(readId, true);
+            reads->setPalindromicFlag(readId, true);
 
         }
     }
@@ -1042,11 +1042,11 @@ bool Assembler::suppressAlignment(
     // don't suppress the alignment.
     // Check the channel first for efficiency,
     // so we can return faster in most cases.
-    const auto ch0 = reads.getMetaData(readId0, "ch");
+    const auto ch0 = reads->getMetaData(readId0, "ch");
     if(ch0.empty()) {
         return false;
     }
-    const auto ch1 = reads.getMetaData(readId1, "ch");
+    const auto ch1 = reads->getMetaData(readId1, "ch");
     if(ch1.empty()) {
         return false;
     }
@@ -1058,11 +1058,11 @@ bool Assembler::suppressAlignment(
 
     // If the sampleid meta data fields of the two reads are missing or different,
     // don't suppress the alignment.
-    const auto sampleid0 = reads.getMetaData(readId0, "sampleid");
+    const auto sampleid0 = reads->getMetaData(readId0, "sampleid");
     if(sampleid0.empty()) {
         return false;
     }
-    const auto sampleid1 = reads.getMetaData(readId1, "sampleid");
+    const auto sampleid1 = reads->getMetaData(readId1, "sampleid");
     if(sampleid1.empty()) {
         return false;
     }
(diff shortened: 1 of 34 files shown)


src/ReadLoader.cpp
```diff
@@ -66,6 +66,12 @@ ReadLoader::ReadLoader(
         "Supported file extensions are .fasta, .fa, .FASTA, .FA.");
 }
 
+
+// This is required because unique_ptr has a type completeness requirement
+// during destruction.
+ReadLoader::~ReadLoader() = default;
```
bagashe, line 72: https://stackoverflow.com/questions/6012157/is-stdunique-ptrt-required-to-know-the-full-definition-of-t/6089065#6089065 has more information on how shared_ptr and unique_ptr differ for type completeness requirements.

src/AssemblerReads.cpp
```diff
@@ -183,14 +184,90 @@ void Assembler::writeReadsSummary()
...
+        csv << reads->getMetaData(readId, "start_time") << ",";
 
         // End of the line for this read.
         csv << "\n";
-     }
+    }
+}
+
+
+void Assembler::adjustCoverage(uint64_t minReadLength, uint64_t desiredCoverage) {
```
paoloczi, line 199: The first argument `minReadLength` is not really needed by this function. It is only used for an error message and an assertion. I would remove this argument from this function as it does not logically belong here - this function does not care about its value. It only cares about `desiredCoverage`. I suggest removing this argument and adding as a return value the `newMinReadLength` (or 0 if insufficient coverage). The error message and assertion can then be done in the caller.

src/AssemblerReads.cpp
```diff
@@ -183,14 +184,90 @@ void Assembler::writeReadsSummary()
...
+            "With a Reads.minReadLength of " + to_string(minReadLength) + ","
+            "the total available coverage (" + to_string(cumulativeBaseCount) +
+            ") is lesser than the desired coverage (" + to_string(desiredCoverage) +
+            "). Try reducing Reads.minReadLength if appropriate or get more coverage."
+        );
+    }
+
+    uint64_t newMinReadLength = std::numeric_limits<uint64_t>::max();
+
+    const auto& binnedHistogram = reads->getBinnedHistogram();
```
paoloczi, line 223: Why do it on the binned histogram instead of the complete histogram, which is also available? As written, it will always use a multiple of 1 Kb as the new read length cutoff.

src/AssemblerReads.cpp
```diff
@@ -183,14 +184,90 @@ void Assembler::writeReadsSummary()
...
+
+
+void Assembler::adjustCoverage(uint64_t minReadLength, uint64_t desiredCoverage) {
+    if (desiredCoverage == 0) {
+        // Not specified. Skip.
+        return;
+    }
+
+    // Need to compute histograms to adjust coverage.
+    reads->computeReadLengthHistogram();
```
paoloczi, line 206: With the code as written, the read length histogram `csv` files refer to the reads actually used by the assembler. This is consistent with current behavior. Perhaps it makes sense to also write separate `csv` files including all reads longer than `minReadLength`? To do this, we could call `Assembler::histogramReadLength` with a different `fileName` (e. g. `ExtendedReadLengthHistogram`) before calling `Assembler::adjustCoverage`. In that case, this call to `Reads::computeReadLengthHistogram` could be removed. 

Both kinds of histogram files would be useful. The ones including all reads longer than `minReadLength` could be used to study possible effects of different values of `minReadLength`.

But if `desiredCoverage` is 0, only one set of csv files should be written, like we do now.

src/AssemblerReads.cpp
```diff
@@ -183,14 +184,90 @@ void Assembler::writeReadsSummary()
...
+        if (cumulativeBaseCount > desiredCoverage) {
+            cumulativeBaseCount -= baseCount;
+            continue;
+        }
+
+        newMinReadLength = max(uint64_t(0), bin - 1) * 1000;
+        break;
+    }
+
+    if (newMinReadLength == std::numeric_limits<uint64_t>::max()) {
```
paoloczi, line 237: Using the complete histogram, instead of the binned one, would probably simplify this logic.

src/AssemblerReads.cpp
```diff
@@ -183,14 +184,90 @@ void Assembler::writeReadsSummary()
...
+    );
+
+    newReads->copyDataForReadsLongerThan(
+        getReads(),
+        newMinReadLength
+    );
+
+    reads->remove();
+    
+    reads.swap(newReads);
```
paoloczi, line 269: All we need is `reads = newReads`, right? The call to `swap` also assigns a new value to `newReads`, but that is superfluous as `newReads` goes immediately out of scope. So the call to `swap` seems confusing to me.

Another alternative would be do to without the `unique_ptr` altogether, and have `class Reads` provide a `swap` member function similar in spirit to `std::vector::swap()`, which would just swap the pointers between the `*this` and the `Reads` passed as argument.

src/Reads.cpp
```diff
@@ -37,6 +37,55 @@ void Reads::access(
...
+void Reads::copyDataForReadsLongerThan(const Reads& rhs, uint64_t newMinReadLength) {
+    for(ReadId id = 0; id < rhs.readCount(); id++) {
+        const auto len = rhs.getReadRawSequenceLength(id);
+        if (len > newMinReadLength) {
+            // Copy over stuff.
+            readNames.appendVector(rhs.readNames.begin(id), rhs.readNames.end(id));
+            readMetaData.appendVector(rhs.readMetaData.begin(id), rhs.readMetaData.end(id));
+            reads.append(rhs.reads[id]);
+            const uint64_t j = readRepeatCounts.size();
+            readRepeatCounts.appendVector(rhs.readRepeatCounts.size(id));
```
paoloczi, line 64: Could this also use a pattern similar to the above calls to `appendVector`, without using a `copy`? Any special reason why this one is done differently?

src/AssemblerReads.cpp
```diff
@@ -183,14 +184,90 @@ void Assembler::writeReadsSummary()
...
+    if (desiredCoverage > cumulativeBaseCount) {
+        throw runtime_error(
+            "With a Reads.minReadLength of " + to_string(minReadLength) + ","
+            "the total available coverage (" + to_string(cumulativeBaseCount) +
+            ") is lesser than the desired coverage (" + to_string(desiredCoverage) +
+            "). Try reducing Reads.minReadLength if appropriate or get more coverage."
+        );
+    }
+
+    uint64_t newMinReadLength = std::numeric_limits<uint64_t>::max();
```
paoloczi, line 221: The new read length cutoff is important enough that it should be stored in `AssemblerInfo` and written to `AssemblySummary.html` and `AssemblySummary.json`. I would store and write the actual read length cutoff used by the assembler - whether specified as `--Reads.minReadLength` or computed here.

src/AssemblerReads.cpp
paoloczi: In the latter alternative, all changes caused by `reads` becoming a `unique_ptr` would no longer be needed.

src/AssemblerAlign.cpp
```diff
@@ -527,7 +527,7 @@ void Assembler::accessCompressedAlignments()
 void Assembler::computeAlignmentTable()
 {
     alignmentTable.createNew(largeDataName("AlignmentTable"), largeDataPageSize);
-    alignmentTable.beginPass1(ReadId(2 * reads.readCount()));
+    alignmentTable.beginPass1(ReadId(2 * reads->readCount()));
```
paoloczi, line 530: At another time, we should perhaps add a member function `Assembler::readCount()`. But this is for another day.

src/AssemblerOptions.cpp
```diff
@@ -201,6 +203,11 @@ void AssemblerOptions::addConfigurableOptions()
         default_value(10000),
         "Read length cutoff. Shorter reads are discarded.")
 
+        ("Reads.desiredCoverage",
+        value<string>(&readsOptions.desiredCoverageString)->
+        default_value("0"),
+        "Desired coverage as a number of bases. Coverage will be reduced to this if necessary.")
```
paoloczi, line 209: I will update the documentation and `shasta.conf` accordingly, and possibly expand the message a bit.

src/AssemblerOptions.cpp
```diff
@@ -753,7 +761,30 @@ void AssemblerOptions::MarkerGraphOptions::parseSimplifyMaxLength()
...
+    desiredCoverage = std::stoull(desiredCoverageString);
 
+    size_t pos = desiredCoverageString.find_last_of("0123456789");
+    if (pos == desiredCoverageString.size() - 1) {
+        // Raw number of bases specified.
+        return;
+    }
+    pos++;
+    
+    while (isspace(desiredCoverageString[pos])) {
```
paoloczi, line 774: We should not allow white space between the number and the multiplier.

src/AssemblerOptions.cpp
```diff
@@ -753,7 +761,30 @@ void AssemblerOptions::MarkerGraphOptions::parseSimplifyMaxLength()
 
 }
 
+void AssemblerOptions::ReadsOptions::parseDesiredCoverageString() {
+    desiredCoverage = std::stoull(desiredCoverageString);
 
+    size_t pos = desiredCoverageString.find_last_of("0123456789");
```
paoloczi, line 767: You could also use a second argument to `stoull` to get the position of the first non-numeric character. That would also be more robust because the code as written would not deal well with a string like "18G2". It would just interpret it as "18".

src/AssemblerReads.cpp
bagashe: Yeah. I'll simply do `reads = std::move(newReads);` instead.

src/Reads.cpp
bagashe: May be. I copy-pasted this code without thinking about potentially rewriting it. This is similar to the code we use to transfer data from per-thread datastructures into the main ones.

src/AssemblerOptions.cpp
bagashe: Agreed. For this to work, the parameter itself would need to be wrapped in `"`, making it a very unlikely scenario.

src/AssemblerOptions.cpp
bagashe: Nice! I like this.

src/LongBaseSequence.cpp
```diff
@@ -16,6 +16,14 @@ void LongBaseSequences::createNew(
...
         data.createNew(name + "-Bases", pageSize);
     }
+    this->name = name;
+}
+
+void LongBaseSequences::rename(const string& name) {
+    if (!name.empty()) {
+        baseCount.rename(name + "-BaseCount");
+        data.rename(name + "-Bases");
+    }
```
paoloczi, line 26: Add an `else` to assert that the existing names are also empty?

src/ReadLoader.cpp
```diff
@@ -663,7 +669,7 @@ void ReadLoader::processCompressedRunnieFile()
     // Use multithreaded code to store the reads.
     setupLoadBalancing(readIdTable.size(), 1000);
     runThreads(&ReadLoader::processCompressedRunnieFileThreadFunction, threadCount);
-    compressedRunnieReader = 0;
+    // compressedRunnieReader = 0;
```
paoloczi, line 672: We still want the equivalent of this for `unique_ptr` (`reset`?), to make sure the reader gets destroyed.

src/Reads.cpp
```diff
@@ -37,6 +37,55 @@ void Reads::access(
...
+
+    reads.rename(readsDataName + suffix);
+    readNames.rename(readNamesDataName + suffix);
+    readMetaData.rename(readMetaDataDataName + suffix);
+    readRepeatCounts.rename(readRepeatCountsDataName + suffix);
+    readFlags.rename(readFlagsDataName + suffix);
+}
+
+
+void Reads::copyDataForReadsLongerThan(const Reads& rhs, uint64_t newMinReadLength) {
```
paoloczi, line 55: This is probably very fast, but I would be curious to know how long it would typically take for a human assembly.

bagashe: I'm going to run a human genome assembly today. Planning to kill it once the reads are processed.

src/ReadLoader.cpp
bagashe: It gets destroyed when it goes out of scope, few lines below this, where the method ends.

paoloczi: It is a data member, so it does not go out of scope at the end of this function.

bagashe: You're right. This is a class data member.

src/Reads.cpp
bagashe: See updated PR description.

src/AssemblerOptions.cpp
```diff
@@ -594,6 +601,7 @@ void AssemblerOptions::ReadsOptions::write(ostream& s) const
 {
     s << "[Reads]\n";
     s << "minReadLength = " << minReadLength << "\n";
+    s << "desiredCoverage = " << desiredCoverage << "\n";
```
paoloczi, line 604: We should write `desiredCoverageString` here, to exactly reflect what we got from the user when running the assembly. I will still merge the PR - you can make this change later.

src/AssemblerReads.cpp
```diff
@@ -183,14 +186,79 @@ void Assembler::writeReadsSummary()
...
+    for (uint64_t length = 0; length < histogram.size(); length++) {
+        const uint64_t frequency = histogram[length];
+        if (frequency) {
+            const uint64_t baseCount = frequency * length;
+            if (cumulativeBaseCount > desiredCoverage) {
+                cumulativeBaseCount -= baseCount;
+                continue;
+            }
+
+            assemblerInfo->minReadLength = length;
```
paoloczi, line 223: The loop ends when `cumulativeBaseCount` drops below `desiredCoverage` (equality included). So in most cases actual coverage will be a bit below what the user specified. I think it would be better instead to set `minReadLength` to the lowest value such that `cumulativeBaseCount` was greater or equal to `desiredCoverage`. That way actual coverage would be a bit more than the user specified (or the same), rather than a bit less. 

You can make this change in a separate PR.

paoloczi: With that change, assuming there is at least one read to begin with, you would no longer have to deal with the case `minReadLength ==std::numeric_limits<uint64_t>::max()` in `main`.

PR #181, opened 2020-08-13
Minor improvements/fixes in the HTTP Server

bagashe:
(Edited)
If the docs directory is available at the expected location, the docs links will be rendered in the HTTP server. Otherwise they won't.

### DIFF
--- a/src/AssemblerHttpServer-Alignments.cpp
+++ b/src/AssemblerHttpServer-Alignments.cpp
@@ -1853,11 +1853,16 @@ void Assembler::exploreAlignmentGraph(
     double timeout= 30;
     getParameterValue(request, "timeout", timeout);
 
-
+    string readGraphHeading;
+    if (httpServerData.docsDirectory.empty()) {
+        readGraphHeading = "<h3>Display a local subgraph of the global alignment graph</h3>";
+    } else {
+        readGraphHeading =
+            "<h3>Display a local subgraph of the <a href='docs/ComputationalMethods.html#ReadGraph'>global alignment graph</a></h3>";
+    }
 
     // Write the form.
-    html <<
-        "<h3>Display a local subgraph of the <a href='docs/ReadGraph.html'>global alignment graph</a></h3>"
+    html << readGraphHeading <<
         "<form>"
 
         "<table>"
--- a/src/AssemblerHttpServer.cpp
+++ b/src/AssemblerHttpServer.cpp
@@ -55,6 +55,7 @@ void Assembler::processRequest(
 {
     // Process a documentation request.
     const string& keyword = request.front();
+    
     if(keyword.size()>6 && keyword.substr(0, 6)=="/docs/") {
 
         // Extract the file name.
@@ -226,7 +227,6 @@ void Assembler::writeHtmlBegin(ostream& html, bool navigation) const
         "<!DOCTYPE html>"
         "<html>"
         "<head>"
-        "<link rel=icon href=docs/CZI-new-logo.png />"
         "<meta charset='UTF-8'>"
         "<title>Shasta assembler</title>";
     writeStyle(html);
@@ -325,9 +325,12 @@ void Assembler::writeNavigation(ostream& html) const
         {"Assembly graph edges support", "exploreAssemblyGraphEdgesSupport"},
         {"Compressed assembly graph", "exploreCompressedAssemblyGraph"},
         });
-    writeNavigation(html, "Help", {
-        {"Documentation", "docs/index.html"},
-        });
+    
+    if (!httpServerData.docsDirectory.empty()) {
+        writeNavigation(html, "Help", {
+            {"Documentation", "docs/index.html"},
+            });
+    }
 
     html << "</ul>";
 }
--- a/src/filesystem.cpp
+++ b/src/filesystem.cpp
@@ -256,3 +256,33 @@ string shasta::filesystem::getAbsolutePath(const string& path)
     ::realpath(path.c_str(), buffer.data());
     return string(buffer.data());
 }
+
+
+string shasta::filesystem::executablePath() {
+    string path;
+    vector<char> buf(1024, 0);
+#ifdef __linux__
+    size_t bufSize = buf.size();
+    ssize_t bytesRead = readlink("/proc/self/exe", &buf[0], bufSize);
+    if (bytesRead < 0) {
+        throw runtime_error("Could not read path of executable.");
+    }
+    path = string(&buf[0], bytesRead);
+#else
+// _NSGetExecutablePath() does not compile. Including <mach-o/dyld.h> causes
+// compilation errors. Keeping the code around, in case we want to make this work on macOS
+// at a later time.
+// #include <mach-o/dyld.h>
+//     uint32_t size = static_cast<uint32_t>(buf.size());
+//     int result = _NSGetExecutablePath(&buf[0], &size);
+//     if (result < 0) {
+//         throw runtime_error("Could not read path of executable.");
+//     }
+//     path = string(&buf[0], size);
+#endif
+    return path;
+}
+
+string shasta::filesystem::getParentDirectoryPath(const string& path) {
+    return path.substr(0, path.find_last_of('/'));
+}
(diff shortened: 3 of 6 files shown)


paoloczi:
Your comment says "ComputeAllAlignments will no longer crash the server if there is a SeqAn error". Was it really crashing the server? If so, there is another problem, because the server should have a `catch` block that should trap the error, write it to html, then close the connection and go back to the connection loop. If this was not happening, there is another problem that also needs to be looked at.

src/AssemblerHttpServer-Alignments.cpp
```diff
@@ -1520,23 +1520,29 @@ void Assembler::computeAllAlignmentsThreadFunction(size_t threadId)
...
+                    } else if (method == 3) {
+                        alignOrientedReads3(
+                            orientedReadId0, orientedReadId1,
+                            matchScore, mismatchScore, gapScore,
+                            downsamplingFactor, bandExtend, maxBand,
+                            alignment, alignmentInfo);
+                    } else {
+                        SHASTA_ASSERT(0);
+                    }
+                } catch (const std::exception& e) {
```
paoloczi, line 1542: Also add a catch block for a non-`std` exception (`catch(...)`), like in `computeAlignments`? Just in case SeqAn throws a custom exception class that is not derived from `std::exception`.

src/AssemblerHttpServer-Alignments.cpp
```diff
@@ -1520,23 +1520,29 @@ void Assembler::computeAllAlignmentsThreadFunction(size_t threadId)
...
+                    cout << e.what() << " for reads " << orientedReadId0 << " and " << orientedReadId1
+                        << ", with bandExtend = " << bandExtend << " and maxBand = " << maxBand << endl;
```
paoloczi, line 1544: No need to write `bandExtend` and `maxBand` but it would be useful to have the band parameters passed to SeqAn. Those would have to be added to the exception message.

src/AssemblerHttpServer-Alignments.cpp
```diff
@@ -1590,7 +1596,7 @@ void Assembler::exploreAlignmentGraph(
 
     // Write the form.
     html <<
-        "<h3>Display a local subgraph of the <a href='docs/ReadGraph.html'>global alignment graph</a></h3>"
+        "<h3>Display a local subgraph of the global alignment graph</h3>"
```
paoloczi, line 1599: In an ideal world, it would be useful to have links to the documentation from the http server. We could even have documentation specific to the http server that explains how to use a page, for example. There are two scenarios where this currently works (or, at least, should work): when the server runs from the `install` directory (possibly obtained from a `tar` artifact), and when the server runs from the `AppImage`. We are currently not distributing the `AppImage`, but we could consider switching to that, instead of the static executable.

We can think about this in the future, but I agree on removing this link for now.

src/AssemblerHttpServer.cpp
```diff
@@ -55,35 +55,7 @@ void Assembler::processRequest(
...
-        }
-
-        // Send it to html.
-        html << "\r\n" << file.rdbuf();
-        return;
-    }
-
-
-
+
```
paoloczi, line 58: Perhaps leave this in under `#ifdef` or commented out, to make it easier to reinstate if we decide to do that in the future?

src/AssemblerHttpServer.cpp
```diff
@@ -226,7 +198,7 @@ void Assembler::writeHtmlBegin(ostream& html, bool navigation) const
         "<!DOCTYPE html>"
         "<html>"
         "<head>"
-        "<link rel=icon href=docs/CZI-new-logo.png />"
+        "<link rel=icon href=https://chanzuckerberg.github.io/shasta/CZI-new-logo.png />"
```
paoloczi, line 201: I don't want to add a reliance on GitHub Pages, for various reasons that we can talk about. I would completely remove the link, perhaps leaving it in, commented out, to make it easier to put it back if we decide to do that.

src/AssemblerHttpServer.cpp
```diff
@@ -326,7 +298,7 @@ void Assembler::writeNavigation(ostream& html) const
         {"Compressed assembly graph", "exploreCompressedAssemblyGraph"},
         });
     writeNavigation(html, "Help", {
-        {"Documentation", "docs/index.html"},
+        {"Documentation", "https://chanzuckerberg.github.io/shasta/index.html"},
```
paoloczi, line 301: Let's take this out for now (leave in, commented out?). As is, this would point to a version of the documentation that is generally not consistent with the executable being run. Also, I don't want to introduce a reliance on GitHub Pages (per my other comment above).

Alternatively, we could leave this in, conditional to the `docs` directory being found in the expected location and containing `index.html`. I think it used to be that way at some point a long time ago. The check might have gone away as part of a restructuring of the navigation code.

bagashe:
Accidentally closed. Re-opening for iteration.

src/AssemblerHttpServer-Alignments.cpp
bagashe: `AppImage` has come up twice in our conversations. I need to look into what that is :)

src/AssemblerHttpServer.cpp
bagashe: We can always put the CZI logo in an s3 bucket and link to it. I will just comment it out for now.

src/AssemblerHttpServer.cpp
bagashe: I will comment it out for now.

src/AssemblerHttpServer-Alignments.cpp
paoloczi: You will find lots of documentation and information on it. An AppImage contains the executable plus all the dynamic libraries it requires plus all the support files you want (e. g. for us `docs`, `conf`) and stores them in compressed form in a single file. At execution time, all of those files are mounted under the covers in a filesystem (without decompressing them) and made available to the executable. For example, the way it is built now for Shasta, the executable knows that the `docs` directory is always available at `../docs` relative to the location of the executable, just like when the executable runs form the `shasta-install/bin` directory.

The AppImage is supposed to be portable across Linux systems. However there is a guideline that says you should build it on the oldest system you want to support, which could be problematic for us. I am not sure it this is a strict requirement though.

bagashe:
If the docs directory is available at the expected location, the docs links will be rendered in the HTTP server. Otherwise they won't.
This only works for Linux at the moment. I couldn't get the macOS version to compile on Github actions. Since I don't have a mac, its hard to test and iterate.

src/filesystem.cpp
```diff
@@ -256,3 +256,33 @@ string shasta::filesystem::getAbsolutePath(const string& path)
     ::realpath(path.c_str(), buffer.data());
     return string(buffer.data());
 }
+
+
+string shasta::filesystem::executablePath() {
+    string path;
+    vector<char> buf(1024, 0);
```
paoloczi, line 263: Consider changing 1024 to `PATH_MAX`, even though `PATH_MAX` is flawed in some ways?

src/AssemblerHttpServer.cpp
```diff
@@ -325,9 +325,12 @@ void Assembler::writeNavigation(ostream& html) const
         {"Assembly graph edges support", "exploreAssemblyGraphEdgesSupport"},
         {"Compressed assembly graph", "exploreCompressedAssemblyGraph"},
         });
-    writeNavigation(html, "Help", {
-        {"Documentation", "docs/index.html"},
-        });
+    
+    if (!httpServerData.docsDirectory.empty()) {
```
paoloczi, line 329: I would like to gradually switch away from `!`, `&&`, `||` in favor of keywords `not, and, or`. But it is not essential to make this change here.

src/filesystem.cpp
```diff
@@ -256,3 +256,33 @@ string shasta::filesystem::getAbsolutePath(const string& path)
...
+//     if (result < 0) {
+//         throw runtime_error("Could not read path of executable.");
+//     }
+//     path = string(&buf[0], size);
+#endif
+    return path;
+}
+
+string shasta::filesystem::getParentDirectoryPath(const string& path) {
+    return path.substr(0, path.find_last_of('/'));
```
paoloczi, line 287: This assumes tat the path passed in is absolute. Check for it?

bagashe: Good point. Will do.

PR #182, opened 2020-08-17
Best effort auto-selection of MinHash.minBucketSize & MinHash.maxBucketSize if values are not provided.

bagashe:
It's tricky picking a good value for `minBucketSize` & `maxBucketSize` for new data/genome. If `minBucketSize` is set too low, then poor quality alignment candidates are let through. If it is set too high, then Shasta misses out on some good alignment candidates. One of the first pieces of feedback when someone assembles a genome using Shasta is to adjust this value so as to get better results. Turns out that we can use simple heuristics to find a **_not-terrible starting value_**.

### Test Plan
Ran an E-Coli assembly before and after this change with `MinHash.minBucketSize = 5`. Verified that the relevant csv files were identical. 

Ran an E-Coli assembly by passing in `MinHash.minBucketSize = 0`. Verified that a new value of minBucketSize was computed and used for each iteration.

Ran an HG002 assembly by passing in `MinHash.minBucketSize = 0` and `MinHash.maxBucketSize = 0`. Verified that reasonable values were selected for both in every min hash iteration.

Tested only LowHash0.

### DIFF
--- a/src/LowHash0.cpp
+++ b/src/LowHash0.cpp
@@ -368,9 +368,6 @@ void LowHash0::pass1ThreadFunction(size_t threadId)
 }
 
 void LowHash0::computeBucketSizesIfNotProvided() {
-    const double minBucketCutOff = 0.65;
-    const double maxBucketCutOff = 0.9;
-    
     uint64_t cumulativeFeatureCount = 0;
     vector<uint64_t> histogram(bucketHistogram.size(), 0);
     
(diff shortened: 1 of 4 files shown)


paoloczi:
In a "healthy" assembly, the minimum and maximum in the histogram are much more pronounced than in the illustration. This has no implications on the PR though.

src/LowHash0.cpp
```diff
@@ -356,6 +367,44 @@ void LowHash0::pass1ThreadFunction(size_t threadId)
...
+    } catch (const PeakFinderException& e) {
+        uint64_t prevFeatureCount = 0;
+        for (uint64_t bucketSize = 0; bucketSize < histogram.size(); bucketSize++) {
+            if (prevFeatureCount == 0) {
+                prevFeatureCount = histogram[bucketSize];
+                continue;
+            }
+
+            double ratio = double(histogram[bucketSize]) / double(prevFeatureCount);
+            if (ratio > 0.8) {
```
paoloczi, line 393: Do you mean to check for `ratio < 0.8`?

bagashe: No. I want the drop to be less than 20% - the reduced slope indicating that we're close to bottoming out. May be a more readable version of this would be ...
```
double drop = (prevFeatureCount - histogram[bucketSize])/prevFeatureCount;
if (drop <= 0.2) {
  ...
}
```

src/LowHash0.cpp
```diff
@@ -573,7 +617,9 @@ void LowHash0::computeBucketHistogram()
     for(const vector<uint64_t>& histogram: threadBucketHistogram) {
         largestBucketSize = max(largestBucketSize, uint64_t(histogram.size()));
     }
-    vector<uint64_t> bucketHistogram(largestBucketSize, 0);
+    
+    bucketHistogram.resize(largestBucketSize, 0);
```
paoloczi, line 621: This only sets to zero any values that were not previously present. To make sure the entire vector is zeroed, you need to also use `std::fill`.

src/LowHash0.cpp
paoloczi: I see. Honestly, I am not sure how robust this will be. I am not sure I want to offer this to users as an alternative for dealing with setting the value themselves. Even if we do it with a big disclaimer, people could start using it indiscriminately.

PR #183, opened 2020-08-17
Handle empty downsampled sequences before attempting a SeqAn alignment.

bagashe:
Also handle SeqAn error gracefully when computing all alignments in the HTTP server.

### Test Plan
For E-Coli data set, this error condition was triggered for the read pair (1000, 4092). So I was able to test by computing the alignment between those two reads in the HTTP server.

### DIFF
--- a/src/AssemblerAlign3.cpp
+++ b/src/AssemblerAlign3.cpp
@@ -95,8 +95,7 @@ void Assembler::alignOrientedReads3(
 
     }
 
-    if ((downsampledSequences[0].data_begin == downsampledSequences[0].data_end) or
-        (downsampledSequences[1].data_begin == downsampledSequences[1].data_end)) {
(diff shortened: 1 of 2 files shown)


src/AssemblerAlign3.cpp
```diff
@@ -95,7 +95,14 @@ void Assembler::alignOrientedReads3(
 
     }
 
-
+    if ((downsampledSequences[0].data_begin == downsampledSequences[0].data_end) or
```
paoloczi, line 98: I would avoid using these undocumented SeqAn fields. Every time we add something to `downsampledSequences` we also add it to `downsampledMarkers`, so I would replace this with 

```
if(downsampledMarkers[0].empty() or downsampledMarkers[1].empty())
```

PR #186, opened 2020-08-27
GUI improvements for "Assess Alignments" page

rlorigro:
Some changes/additions to the browser page:

1. The output of each iteration is now stored in a table (even when verbose display is on)
2. Histogram max and bin size are now configurable
3. Clarification of empty fields in min/max read length
4. Random sampling has been modified to be non-deterministic (depending on whether you believe in universal determinism)

**Example 1:**
image

**Example 2 and 3:**
image


rlorigro:
I think I would like to add an option to sample only from the read graph as well. Maybe that will be the next PR for this browser page

PR #188, opened 2020-08-31
Limited automation of Align parameters

rlorigro:
This adds a simple method for evaluating the alignment drift, skip, trim, alignedFraction, and markerCount during run time, and choosing cutoffs based on a percentile. Testing on 3 basecallers and 1 ultralong dataset confirms that performance in terms of continuity and total assembled length is equal to or better than manually selected parameters.

For future development, more CSV output files are created when running with `ReadGraph.creationMethod 2`. This includes the distribution of all (5) parameters that are currently automated, as well as a list of alignments and their individual stats regarding these parameters.

maxTrim is currently still being evaluated, so perhaps it is best to wait on merging until those results are in.

2 new config files are added, which enable automation:
1. Nanopore-Jun2020-Automation.conf
2. Nanopore-UL-Jun2020-Automation.conf

The major changes to these config files are:
- Substitute `maxSkip`, `maxDrift`, `maxTrim`, `minAlignedMarkerCount`, and `minAlignedFraction` for very permissive values 
- Enable `ReadGraph.CreationMethod 2`
- Enable `MarkerGraph.maxCoverage 0`, which automates marker vertex coverage thresholding using PeakFinder
- Remove `ReadGraph.maxAlignmentCount 12` from the UL config (as a result of testing)

### DIFF
--- a/conf/Nanopore-Jun2020-Automation.conf
+++ b/conf/Nanopore-Jun2020-Automation.conf
@@ -1,0 +1,88 @@
+# DO NOT USE THIS FILE IF YOU HAVE READS CREATED BY A
+# GUPPY VERSION OLDER THAN 3.6.0.
+
+# This file contains Shasta options that, as of June 2020,
+# are known to work with Oxford Nanopore reads under the following 
+# circumstances:
+
+# - Human genome assembly.
+# - Coverage between 40x and 80x. If you have more coverage than that,
+#   you can use option "--Reads.minReadLength" to adjust coverage as desired.
+# - Reads from Guppy 3.6.0 base caller. If you have reads
+#   from an older version of Guppy, use configuration file
+#   Nanopore-Dec2019.conf instead.
+
+# To use this configuration file, specify Shasta option 
+# "--config AbsolutePathToThisFile". 
+# If you specify any conflicting values on the command line,
+# the values specified on the command line take precedence.
+
+# In most cases, for best performance on a large assembly 
+# you will usually also want to use the following options, which 
+# cannot be specified in a configuration file:
+# --memoryMode filesystem
+# --memoryBacking 2M
+# Using these options requires root access.
+
+# For detailed information on all available options see here:
+# https://chanzuckerberg.github.io/shasta/CommandLineOptions.html
+
+# For information on running a small assembly for which 
+# performance is not essential see here:
+# https://chanzuckerberg.github.io/shasta/QuickStart.html
+
+# For more information on running an assembly see here:
+# https://chanzuckerberg.github.io/shasta/Running.html
+
+# For information on optimizing assembly performance see here:
+# https://chanzuckerberg.github.io/shasta/Performance.html
+
+
+
+[Reads]
+# If you have extra coverage, use this option to adjust coverage.
+minReadLength = 10000
+noCache = True
+
+[Kmers]
+# Due to the higher accuracy of Guppy 3.6.0 we use longer
+# markers than usual.
+k = 14
+
+[MinHash]
+minBucketSize = 5
+maxBucketSize = 30
+minFrequency = 5
+
+[Align]
+alignMethod = 3
+downsamplingFactor = 0.05
+matchScore = 6
+sameChannelReadAlignment.suppressDeltaThreshold = 30
+
+# The following Align parameters are set to very permissive values to allow the majority of alignments
+# to be assessed during the initial stage of automatic alignment parameter selection
+maxSkip = 100
+maxDrift = 100
+maxTrim = 100
+minAlignedMarkerCount = 10
+minAlignedFraction = 0.1
+
+[ReadGraph]
+# This method uses the observed distribution of alignment stats to choose a cutoff for
+# maxSkip, maxDrift, maxTrim, minAlignedMarkerCount, and minAlignedFraction
+creationMethod = 2
+
+[MarkerGraph]
+simplifyMaxLength = 10,100,1000,10000,100000
+refineThreshold = 6
+crossEdgeCoverageThreshold = 3
+
+# Automatically determine this using PeakFinder
+minCoverage = 0
+
+[Assembly]
+consensusCaller = Bayesian:guppy-3.6.0-a
+detangleMethod = 1
+
+
--- a/conf/Nanopore-UL-Jun2020-Automation.conf
+++ b/conf/Nanopore-UL-Jun2020-Automation.conf
@@ -1,0 +1,87 @@
+# DO NOT USE THIS FILE IF YOU HAVE READS CREATED BY A
+# GUPPY VERSION OLDER THAN 3.6.0.
+
+# This file contains Shasta options that, as of June 2020,
+# are known to work with Ultra-Long (UL) Oxford Nanopore reads 
+# under the following circumstances:
+
+# - Human genome assembly.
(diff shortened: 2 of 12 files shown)


conf/Nanopore-Jun2020-Automation.conf
```diff
@@ -0,0 +1,88 @@
+# DO NOT USE THIS FILE IF YOU HAVE READS CREATED BY A
+# GUPPY VERSION OLDER THAN 3.6.0.
+
+# This file contains Shasta options that, as of June 2020,
```
paoloczi, line 4: Let's rename this file `Nanopore-Sep2020.conf`, without the "Automation" suffix. Instead, put a well visible comment near the top of the file to describe what it does. I plan to add a section on `ReadGraph.creationMethod 2` to the documentation.

conf/Nanopore-UL-Jun2020-Automation.conf
```diff
@@ -0,0 +1,87 @@
+# DO NOT USE THIS FILE IF YOU HAVE READS CREATED BY A
+# GUPPY VERSION OLDER THAN 3.6.0.
+
+# This file contains Shasta options that, as of June 2020,
```
paoloczi, line 4: Same comment here regarding the file name.

scripts/CreateReadGraph2.py
```diff
@@ -14,7 +14,6 @@
 
 # Create the global read graph.
 a.createReadGraph2(
-    maxAlignmentCount = int(config['ReadGraph']['maxAlignmentCount']),
-    maxTrim = int(config['Align']['maxTrim']))
+    maxAlignmentCount = int(config['ReadGraph']['maxAlignmentCount']))
```
paoloczi, line 17: I will add the other parameters.

src/AssemblerReadGraph2.cpp
```diff
@@ -4,11 +4,143 @@ using namespace shasta;
...
 void Assembler::createReadGraph2(
-    uint32_t maxAlignmentCount,
-    uint32_t maxTrim)
+    uint32_t maxAlignmentCount)
 {
-    // This boilerplate code creates the read graph using
-    // all available alignments.
-    vector<bool> keepAlignment(alignmentData.size(), true);
+    // Percentile cutoffs for automated parameters
+    double largePercentile = 0.12;
```
paoloczi, line 10: Are these the only two parameters? Do we the flexibility to allow need different percentiles for different quantities?

src/AssemblerReadGraph2.cpp
```diff
@@ -4,11 +4,143 @@ using namespace shasta;
...
+    double smallPercentile = 0.015;
+
+    vector<bool> keepAlignment(alignmentData.size(), false);
+
+    // Initialize histograms for measuring alignedFraction, markerCount, maxDrift, and maxSkip distributions
+    Histogram2 alignedFractionHistogram(0, 1, 100);
+    Histogram2 markerCountHistogram(0, 3000, 300);
```
paoloczi, line 17: For long reads the maximum of 3000 might be insufficient. Consider modifying `Histogram2` so you no longer have to specify the maximum, and it adjusts the length of the vector as needed?

src/AssemblerReadGraph2.cpp
```diff
@@ -4,11 +4,143 @@ using namespace shasta;
...
+    Histogram2 maxDriftHistogram(0, 100, 100);
+    Histogram2 maxSkipHistogram(0, 100, 100);
+    Histogram2 maxTrimHistogram(0, 100, 100);
+
+    ofstream alignmentInfoCsv("alignmentInfo.csv");
```
paoloczi, line 22: I would use a file name beginning with an uppercase letter for consistentcy with other csv files.

paoloczi: Also, avoid "AlignmentInfo" which has another meaning in Shasta.

src/AssemblerReadGraph2.cpp
```diff
@@ -4,11 +4,143 @@ using namespace shasta;
...
+    alignmentInfoCsv << "readId0" << ','
+                     << "readId1" << ','
+                     << "minAlignedFraction" << ','
+                     << "markerCount" << ','
+                     << "maxDrift" << ','
+                     << "maxSkip" << ','
+                     << "trim" << '\n';
+
+    // Sample read pairs by discarding. Not ideal for large number of pairs AND small fractions
```
paoloczi, line 31: This comment seems out of sync with the code.

src/AssemblerReadGraph2.cpp
```diff
@@ -4,11 +4,143 @@ using namespace shasta;
...
+                << info.markerCount << ','
+                << info.maxDrift << ','
+                << info.maxSkip << ','
+                << trim << '\n';
+    }
+
+    // Minimums
+    double alignedFractionThreshold = alignedFractionHistogram.thresholdByCumulativeProportion(largePercentile);
+
+    // MarkerCount is not at all gaussian, so it needs a different percentile.
```
paoloczi, line 55: Add a different percentile for each of the quantities you are histogramming?

src/AssemblerReadGraph2.cpp
rlorigro: I was initially trying not to add more parameters, but the more I think about it, the more it makes sense to have the option to change them independently if needed

src/AssemblerReadGraph2.cpp
rlorigro: This is actually the values contained in AlignmentInfo, written as a csv. Maybe `AlignmentSummaryStats.csv` is better?

paoloczi: Given the contents of this file, I think the name `AlignmentInfo.csv` (without reference to summary or statistics, because this is detailed output) is fine (with uppercase first letter). But like I said in a separate comment, only create it under control of a `const bool debug = false` flag, so you can turn it on easily for debugging if you need to.

rlorigro: ok sounds good, it is pretty large, about 3Gb

src/Histogram.cpp
```diff
@@ -59,12 +71,33 @@ int64_t shasta::Histogram2::findIndex(double x){
 void shasta::Histogram2::update(double x) {
     auto index = findIndex(x);
 
-    if (index >= 0) {
+    if (dynamicBounds){
+        // If the found index is too large, extend the end of the deque
+        if (index > histogram.size()){
+            size_t newBins = size_t(index)-histogram.size();
```
paoloczi, line 77: We have been trying to avoid `size_t` in new code because it has a different size on macOS and its use tends to cause compilation problems on macOS. Please convert to `uint64_t` here and elsewhere.

src/Histogram.hpp
```diff
@@ -59,29 +64,31 @@ class shasta::Histogram : public vector<size_t> {
...
+    size_t binCount;
     const double binSize;
-    vector<uint64_t> histogram;
+    deque<uint64_t> histogram;
     bool unboundedLeft;
     bool unboundedRight;
+    bool dynamicBounds;
 
     /// Methods ///
     Histogram2(
```
paoloczi, line 77: Perhaps add a few lines a comment explaining the functionality of this class and how the various constructor parameters interact with each other?

PR #192, opened 2020-09-23
Updates to Github Actions workflow and MakeRelease.html

bagashe:
### Linux
I think it is a good idea to support building Shasta on all Ubuntu LTS versions that haven't reached their end-of-support yet. So we will keep building on Ubuntu 18.04 LTS for now. If this becomes a problem (for example when we need a later version of `gcc` or some other library), we can consider what to do at the time. 

### MacOS
MacOS 10.14 still has around 20% market share. So I've kept the build step around. We won't collect the build artifact anymore since it won't be part of our release. In general, it is a good idea to support building Shasta on the two latest macOS versions. We can still have a single macOS release if the binary works on both versions of the OS.

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -68,7 +68,6 @@ jobs:
         path: shasta-build/shasta-Ubuntu-18.04.tar
 
 
-
   ubuntu-16-04:
     runs-on: ubuntu-16.04
     steps:
@@ -125,11 +124,6 @@ jobs:
         gunzip tests/TinyTest.fasta.gz
         shasta-build/shasta-install/bin/shasta --Align.alignMethod 3 --input tests/TinyTest.fasta
         ls -l ShastaRun/Assembly.fasta
-    - uses: actions/upload-artifact@master
-      with:
-        name: shasta-macOS-10.14
-        path: shasta-build/shasta-install/bin/shasta
-
 
        
   macOS-10-15:
@@ -156,7 +150,7 @@ jobs:
         ls -l ShastaRun/Assembly.fasta
     - uses: actions/upload-artifact@master
       with:
-        name: shasta-macOS-10.15
+        name: shasta-macOS
         path: shasta-build/shasta-install/bin/shasta
 
 
--- a/docs/MakeRelease.html
+++ b/docs/MakeRelease.html
@@ -32,16 +32,27 @@ change <code>BUILD_ID</code> to <code>Shasta Release X.Y.Z</code>
 in all <code>cmake</code> commands (5 instances at the time of last update of this documentation).
 <li>Commit and push this change to trigger a new build on GitHub Actions.
 Wait for that build to complete.
-<li>Download the 7 artifacts. Unzip them and rename  them to the following:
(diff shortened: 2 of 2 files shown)


.github/workflows/Build.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         path: shasta-build/shasta-Ubuntu-20.04/bin/shasta
     - uses: actions/upload-artifact@master
       with:
-        name: shasta-Ubuntu-20.04.tar
+        name: shasta-Linux.tar
```
paoloczi, line 38: I would keep the Ubuntu version in the name of the tar file because that is not necessarily portable to other Ubuntu versions.

.github/workflows/Build.yml
```diff
@@ -98,7 +94,7 @@ jobs:
         path: shasta-build/shasta-Ubuntu-16.04/bin/shasta
     - uses: actions/upload-artifact@master
       with:
-        name: shasta-Ubuntu-16.04.tar
+        name: shasta-OldLinux.tar
```
paoloczi, line 97: Same here.

docs/MakeRelease.html
```diff
@@ -32,16 +32,26 @@ <h1>Creating a new release</h1>
 in all <code>cmake</code> commands (5 instances at the time of last update of this documentation).
 <li>Commit and push this change to trigger a new build on GitHub Actions.
 Wait for that build to complete.
-<li>Download the 7 artifacts. Unzip them and rename  them to the following:
+<li>Download the 5 artifacts. Unzip them and rename them to the following:
```
paoloczi, line 35: Also the 18.04. Tar files are not portable across Ubuntu versions.

.github/workflows/Build.yml
bagashe: This archive has `bin`, `docs`, `docker` & `conf` directories. `bin` has Python3 scripts and the Shasta binary, which works on Ubuntu 20.04 LTS and Ubuntu 18.04 LTS. 

If we name the executable `shasta-Linux`, then we should name this `shasta-Linux.tar`, no?

docs/MakeRelease.html
bagashe: Not sure I understand this. See my other response.

.github/workflows/Build.yml
paoloczi: No, because the tar file contains the dynamic executable and shared library (Python module), which are not portable across Ubuntu versions. So for example if you try to use the Python API using a 20.04 build on 18.04, it will not work. If you want to use the Python API and you don't want to build the code yourself, you have to download the tar file that was built on the Ubuntu version you are using. For this reason I think we should keep the Ubuntu version in the name of tar files.

For the static executable it's different because it is very portable not just across Ubuntu, but also to other Linux versions.

Conceivably, we could add a portable tar file than contains everything but the dynamic library, the dynamic executable, and the Python API scripts. The `bin` directory would probably only contain the static executable, and perhaps a few Python scripts that don't use `'import shasta`'. This tar file could be given a generic name that does not include the Ubuntu version.

bagashe: Good point about the dynamic libraries.

PR #194, opened 2020-09-24
More parameters for PeakFinder

rlorigro:
This is primarily intended to address a failure in PeakFinder to choose a cutoff when the distribution contains an overwhelmingly large value at x=1. The (temporary?) solution is more configurability for the percent area calculation, allowing x=1 to be omitted.


bagashe:
Would this approach work for the Guppy 4.0 Tomato reads?

PR #198, opened 2020-10-01
Update configs with detangle 2, add oldGuppy

rlorigro:
All Sep2020 configs now have `detangleMethod 2`, and a config has been created for guppy <3.6.0


conf/Nanopore-OldGuppy-Sep2020.conf
```diff
@@ -0,0 +1,87 @@
+# DO NOT USE THIS FILE IF YOU HAVE READS CREATED BY A
```
paoloczi, line 1: This comment is incorrect. Remove or replace.

PR #203, opened 2020-10-07
Post release cleanup

bagashe:
Github Actions now run three builds per commit. _I'm dropping build support for Ubuntu 16, Ubuntu 18 & macOS-10.14._
1. Ubuntu-20-04
2. **Ubuntu-20-04-minimal** (See below for details)
3. macOS-10-15

### Minimal mode
I'm introducing a `minimal` mode for Shasta. In this mode Shasta does not have the embedded HTTP server. The `Install-Prerequisites-Ubuntu.sh` script now supports a `--minimal` flag, which results in a leaner installation of packages, as it skips installing packages necessary only for the HTTP server. It also skips installing packages necessary for Python-C++ bindings.

### Code cleanup
Since we were always building Shasta with embedded HTTP server, several `#ifdef SHASTA_HTTP_SERVER` guards were had ended up in the wrong place. So I needed to move around some code and do a very small refactor, in order to correctly build Shasta, both with and without the embedded HTTP server. Some of the HTML code generation is still part of the the `minimal` build because we need it to generate _AssemblySummary.html_.

On x86_64, SPOA is built with CPU dispatching. On aarch64, SPOA is built without CPU dispatching. This is no longer dependent on the version of Ubuntu. 

I've updated the installation script to install only the boost packages that we need - libboost-system-dev, libboost-program-options-dev, libboost-chrono-dev, libboost-graph-dev.

### Test Plan
1. Built Shasta with HTTP Server support. Assembled _E Coli_ genome and explored it in the server. Verified that it works as before.
2. Built Shasta without HTTP Server support. Assembled _E Coli_ genome and tried to explore it in the server. Verified failure with an error message saying `exploreAccess` isn't supported.


.github/workflows/Build.yml
```diff
@@ -38,93 +38,31 @@ jobs:
...
-    runs-on: ubuntu-18.04
+  ubuntu-20-04-minimal:
+    runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@master
     - name: Install-prerequisites
-      run: sudo ./scripts/InstallPrerequisites-Ubuntu.sh
-    - name: Build-ubuntu-18.04
+      run: sudo ./scripts/InstallPrerequisites-Ubuntu.sh --minimal
+    - name: Build-ubuntu-20.04
```
paoloczi, line 47: This has the same name as the one for the regular Ubuntu build. I am not sure this is a problem, but isn't it better to have a name that include "minimal", to avoid confusion?

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.16)
```
paoloczi, line 1: I would not change this unless it is necessary for some reason. Same comment for the other cmake files.

.github/workflows/Build.yml
bagashe: It's not a problem because it is under a different top-level Github action. I'll update it to avoid confusion.

CMakeLists.txt
bagashe: Cool. Will undo.

PR #207, opened 2020-10-19
Handle O_DIRECT errors gracefully.

bagashe:
If O_DIRECT flag is not supported by the file system, Shasta will fall back to using regular file IO.


bagashe:
This is not ready to be merged yet.

PR #208, opened 2020-10-20
Handle O_DIRECT errors gracefully.

bagashe:
Since O_DIRECT related read failures can happen at file `open` as well `read`, I open and read (a pageSize worth of data) from the file to confirm that O_DIRECT actually works.

### DIFF
--- a/src/ReadLoader.cpp
+++ b/src/ReadLoader.cpp
@@ -10,7 +10,6 @@ using namespace shasta;
 #include "iterator.hpp"
 
 
-
 // Load reads from a fastq or fasta file.
 ReadLoader::ReadLoader(
     const string& fileName,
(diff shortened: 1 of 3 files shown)


src/ReadLoader.cpp
```diff
@@ -507,7 +527,12 @@ void ReadLoader::readFile()
     int flags = O_RDONLY;
 #ifdef __linux__
     if(noCache) {
-        flags |= O_DIRECT;
+        flags = tryDirectIO(fileName);
+        if (flags == (O_RDONLY | O_DIRECT)) {
+            cout << "O_DIRECT flag supported by the file system." << endl;
```
paoloczi, line 532: I don't think we need a message in this case.

src/ReadLoader.cpp
```diff
@@ -507,7 +527,12 @@ void ReadLoader::readFile()
...
+        } else {
+            cout << "O_DIRECT flag not supported by file system." << endl;
```
paoloczi, line 534: I would rephrase in terms closer to the user. Something like "--Reads.noCache was turned off for file ... because it is not supported by the filesystem".

PR #216, opened 2020-12-14
Add candidate indexing and candidate graph

rlorigro:
This sets up the basis for doing graph operations on the candidates. Not sure about the name of the new cpp file.

### DIFF
--- a/src/Assembler.hpp
+++ b/src/Assembler.hpp
@@ -660,8 +660,8 @@ private:
     // They are interpreted with readId0 on strand 0.
     AlignmentCandidates alignmentCandidates;
 
-    // The alignment table stores the AlignmentData that each oriented read is involved in.
-    // Stores, for each OrientedReadId, a vector of indexes into the alignmentData vector.
+    // The alignment table stores the read pair that each oriented read is involved in.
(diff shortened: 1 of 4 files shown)


src/Assembler.hpp
```diff
@@ -659,6 +659,13 @@ class shasta::Assembler :
     // They all have readId0<readId1.
     // They are interpreted with readId0 on strand 0.
     AlignmentCandidates alignmentCandidates;
+
+    // The alignment table stores the AlignmentData that each oriented read is involved in.
```
paoloczi, line 663: The comment needs rephrasing as it makes no reference to `alignmentCandidates`.

paoloczi:
For the name of the new cpp file I suggest `AssemblerAlignmentCandidates.cpp`.

src/AssemblerCandidateGraph.cpp
```diff
@@ -0,0 +1,144 @@
...
+                }
+            }
+        }
+    }
+
+    return true;
+}
+
+
+// Compute alignmentTable from alignmentData.
```
paoloczi, line 82: The comment refers to `alignmentTable`, not `candidateTable`.

PR #218, opened 2020-12-18
Add new browser page for candidate visualization

rlorigro:
This creates a new page under the Alignments tab that enables exploring the candidate graph, and visualizing how candidates are filtered at each step of graph refinement, up to the ReadGraph. It reuses code from The ReadGraph and Alignment visualization pages.

This also adds a method for ordering SVGs based on the templated Graph class. The Edge/Vertex members of Graph are expected to provide a method for finding their numeric ordering.

### DIFF
--- a/src/AssemblerHttpServer-Alignments.cpp
+++ b/src/AssemblerHttpServer-Alignments.cpp
@@ -7,6 +7,7 @@
 #include "AlignmentGraph.hpp"
 #include "Histogram.hpp"
 #include "LocalAlignmentGraph.hpp"
+#include "LocalAlignmentCandidateGraph.hpp"
 #include "platformDependent.hpp"
 #include "PngImage.hpp"
 #include "ReadId.hpp"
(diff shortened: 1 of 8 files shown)


src/AssemblerHttpServer-Alignments.cpp
```diff
@@ -26,6 +27,210 @@ using std::random_device;
 using std::uniform_int_distribution;
 
 
+void Assembler::exploreCandidateGraph(
```
paoloczi, line 30: It would be best to rename this `exploreAlignmentCandidateGraph`.

src/AssemblerHttpServer.cpp
```diff
@@ -218,6 +219,7 @@ void Assembler::writeNavigation(ostream& html) const
         {"Reads", "exploreRead"},
         });
     writeNavigation(html, "Alignments", {
+        {"Candidate graph", "exploreCandidateGraph"},
```
paoloczi, line 222: Change the label to "Alignment candidate graph".

src/writeGraph.hpp
```diff
@@ -43,11 +44,19 @@ namespace shasta {
             const std::map<typename Graph::vertex_descriptor, VertexAttributes>&,
             const std::map<typename Graph::edge_descriptor, EdgeAttributes>&,
             ostream&);
+
+        template<class Graph> void writeOrderedSvg(
```
paoloczi, line 48: Please add some comments here and/or in the `.cpp` file to describe what this does and the meaning of the arguments.

PR #220, opened 2020-12-21
Add color picker to candidate alignment graph visualization

rlorigro:
This adds a color picker that informs the user of the current color configuration and gives a venn diagram visualization of the relationship between Alignment Candidates, Good Alignments, Read Graph edges, and reference overlaps.

image
image


paoloczi:
Nice! I will merge it as is, but when you can please change `writeColorPicker` to be a member function of class `Assembler`. As written, it is a free standing function outside the `shasta` namespace.

PR #228, opened 2021-02-03
Browser updates - reference comparison

rlorigro:
This adds some methods to load a PAF as a graph, and integrates edge labeling into the AlignmentCandidateGraph. Additionally, a feature has been added to the Candidates page which allows the user to only visualize the reference graph edges.

An example of the reference graph, colored by inclusion in reference only (red), candidates (purple), good alignments (blue), and read graph edges (green):
image

### DIFF
--- a/src/ReferenceOverlapMap.hpp
+++ b/src/ReferenceOverlapMap.hpp
@@ -1,0 +1,40 @@
+#ifndef SHASTA_REFERENCEOVERLAPMAP_HPP
+#define SHASTA_REFERENCEOVERLAPMAP_HPP
+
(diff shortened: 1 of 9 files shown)


src/ReferenceOverlapMap.hpp
```diff
@@ -0,0 +1,44 @@
+#ifndef SHASTA_REFERENCEOVERLAPMAP_HPP
+#define SHASTA_REFERENCEOVERLAPMAP_HPP
+
```
paoloczi, line 3: When you can, please add some comments here explaining what this does.

PR #230, opened 2021-02-03
Documentation

rlorigro:
Move + expand the commenting for ReferenceOverlapMap

resolving
https://github.com/chanzuckerberg/shasta/pull/228#discussion_r569478009

### DIFF
--- a/src/ReferenceOverlapMap.hpp
+++ b/src/ReferenceOverlapMap.hpp
@@ -1,6 +1,12 @@
 #ifndef SHASTA_REFERENCEOVERLAPMAP_HPP
 #define SHASTA_REFERENCEOVERLAPMAP_HPP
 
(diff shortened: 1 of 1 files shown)


src/ReferenceOverlapMap.hpp
```diff
@@ -1,6 +1,12 @@
 #ifndef SHASTA_REFERENCEOVERLAPMAP_HPP
 #define SHASTA_REFERENCEOVERLAPMAP_HPP
 
+/// The overlap map is based on a boost interval map. The interval map performs an "aggregation" operation whenever
```
paoloczi, line 4: I meant we need a higher level comment describing how we use these mappings to the reference when displaying the alignment candidate graph. Does not have to be super detailed - a general idea is sufficient.

rlorigro: ok, check out the latest commit

PR #235, opened 2021-03-02
Palindrome quality classifier

rlorigro:
This adds a method to detect palindromes with an obvious deviation in quality in the second strand, as with the picture below:

image

However, some palindromes exist that do not have this quality signature


paoloczi:
Thank you. I will integrate this with the ReadLoader and add an option to turn it on. But it will be some time before I can do it.

PR #237, opened 2021-03-12
Add palindrome q filter

rlorigro:
First commit has been tested on a fastq containing only palindromes with identifiable q scores. The result is that no reads are loaded. Next step is adding config options

### DIFF
--- a/src/PalindromeQuality.cpp
+++ b/src/PalindromeQuality.cpp
@@ -1,5 +1,4 @@
 #include "PalindromeQuality.hpp"
-#include <stdexcept>
 #include "span.hpp"
 #include <cmath>
 
@@ -12,6 +11,7 @@ using std::pow;
 #include <boost/accumulators/statistics.hpp>
 
 using namespace boost::accumulators;
+using namespace shasta;
 
 typedef accumulator_set<float, features<tag::count, tag::mean, tag::variance>> stats_accumulator;
 
@@ -21,7 +21,7 @@ double qualityCharToErrorProbability(char q) {
 }
 
 
-bool shasta::classifyPalindromicQScores(span<char> qualities){
+bool shasta::isPalindromic(span<char> qualities){
     double relativeMeanDifference = 0.09;
     double minimumMean = 0.15;
     double minimumVariance = 0.025;
--- a/src/PalindromeQuality.cpp
+++ b/src/PalindromeQuality.cpp
@@ -26,10 +26,6 @@ bool shasta::isPalindromic(span<char> qualities,
                            double minimumMean,
                            double minimumVariance){
 
-//    double relativeMeanDifference = 0.09;
-//    double minimumMean = 0.15;
-//    double minimumVariance = 0.025;
-
     bool isPalindromic = false;
 
     stats_accumulator leftStats;
--- a/src/PalindromeQuality.cpp
+++ b/src/PalindromeQuality.cpp
@@ -13,7 +13,7 @@ using std::pow;
 using namespace boost::accumulators;
 using namespace shasta;
 
-typedef accumulator_set<float, features<tag::count, tag::mean, tag::variance>> stats_accumulator;
+using stats_accumulator = accumulator_set<float, features<tag::count, tag::mean, tag::variance>>;
 
 
 double qualityCharToErrorProbability(char q) {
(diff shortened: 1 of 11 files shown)


src/ReadLoader.cpp
```diff
@@ -467,6 +467,14 @@ void ReadLoader::processFastqFileThreadFunction(size_t threadId)
             continue;
         }
 
+        // Skip if the q scores have an obvious palindromic characteristic
+        span <char> scores(scoresBegin, scoresEnd);
+        if (isPalindromic(scores)){
```
paoloczi, line 472: You have to add a new category of discarded reads. The category you are incrementing here refers to reads that were discarded because they were too short. When you create that new category, a few things need to happen. Just follow what we do for short reads as a pattern. That will include storing in `AssemblerInfo` and writing to `AssemblySummary.html` and `AssemblySummary.json`.

src/PalindromeQuality.cpp
```diff
@@ -12,6 +11,7 @@ using std::pow;
 #include <boost/accumulators/statistics.hpp>
 
 using namespace boost::accumulators;
+using namespace shasta;
 
 typedef accumulator_set<float, features<tag::count, tag::mean, tag::variance>> stats_accumulator;
```
paoloczi, line 16: I no longer use `typedef`. Instead of `typedef A B`, use `using B = A`. It is clearer and more consistent with most of Shasta code.

src/PalindromeQuality.cpp
```diff
@@ -12,19 +11,20 @@ using std::pow;
...
 double qualityCharToErrorProbability(char q) {
     return pow(10, double(q - 33) / -10.0);
 }
 
 
-bool shasta::classifyPalindromicQScores(span<char> qualities){
-    double relativeMeanDifference = 0.09;
-    double minimumMean = 0.15;
-    double minimumVariance = 0.025;
+bool shasta::isPalindromic(span<char> qualities,
```
paoloczi, line 24: Please format this consistently with other Shasta code. The continuation line should have only one extra tab compared to the preceding line.

src/ReadLoader.cpp
```diff
@@ -386,39 +388,39 @@ void ReadLoader::processFastqFileThreadFunction(size_t threadId)
         SHASTA_ASSERT(scoresEnd > scoresBegin);
 
         // Check the header line.
-        if(headerEnd == headerBegin) {
+        if (headerEnd == headerBegin) {
             throw runtime_error("Empty header line for read at offset " +
-                to_string(headerBegin-fileBegin) + ".");
+                                to_string(headerBegin - fileBegin) + ".");
```
paoloczi, line 393: There are extra tabs here too, and a few lines below.

src/ReadLoader.cpp
```diff
@@ -386,39 +388,39 @@ void ReadLoader::processFastqFileThreadFunction(size_t threadId)
...
+        for (auto it = nameBegin; it != headerEnd; ++it) {
             const char c = *it;
-            if(std::isspace(c)) {
+            if (std::isspace(c)) {
                 break;
             }
             readName.push_back(c);
         }
-        if(readName.empty()) {
+        if (readName.empty()) {
```
paoloczi, line 413: I see lots of reformatting changes, many of which inconsistent with Shasta usage. Your editor is probably making them. Let's not worry about this for this time, but it would be best to avoid this in future PRs.

src/ReadLoader.hpp
```diff
@@ -30,6 +32,7 @@ class shasta::ReadLoader :
         size_t threadCount,
         const string& dataNamePrefix,
         size_t pageSize,
+        const PalindromicReadOptions& palindromicReadOptions,
```
paoloczi, line 35: I would prefer to pass to the `ReadLoader` only the three parameters it needs, rather than the entire `PalindromicReadOptions`. Passing the entire object makes it hard for the reader to keep track of where options are used.

src/ReadLoader.hpp
```diff
@@ -71,9 +79,14 @@ class shasta::ReadLoader :
     const string& dataNamePrefix;
     const size_t pageSize;
 
+    // This object is copied for convenience to specify the following:
```
paoloczi, line 82: See comment above.

paoloczi:
Thank you - this is close. Can you also add entries for the new options in `docs/CommandLineOptions.html`? A terse explanation like the existing ones is sufficient. One day I will add a section to the computational methods describing palindromic read detection, and add links from the options to that section.

PR #239, opened 2021-03-16
Add filter command to main executable

rlorigro:
This enables running shasta's read loader and filtering steps, up to the point of palindrome detection by self-alignment. The reads that remain after filtering are written as pairs of `ReadId,ReadName` in a csv and the assembly is aborted after writing a summary that describes the read stats.

### DIFF
--- a/srcMain/main.cpp
+++ b/srcMain/main.cpp
@@ -124,7 +124,8 @@ void shasta::main::main(int argumentCount, const char** arguments)
 
 
     // Execute the requested command.
-    if(assemblerOptions.commandLineOnlyOptions.command == "assemble") {
+    if(assemblerOptions.commandLineOnlyOptions.command == "assemble" or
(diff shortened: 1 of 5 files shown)


srcMain/main.cpp
```diff
@@ -124,7 +124,8 @@ void shasta::main::main(int argumentCount, const char** arguments)
 
 
     // Execute the requested command.
-    if(assemblerOptions.commandLineOnlyOptions.command == "assemble") {
+    if(assemblerOptions.commandLineOnlyOptions.command == "assemble" or
```
paoloczi, line 127: Please update the `throw` message about 20 lines below to also include `filterReads` as one of the valid commands.

paoloczi: And similarly add the new command to `createBashCompletionScript`.

Issue #229, opened 2021-02-03
Support for macOS 11 (Big Sur)


paoloczi:
Currently the Shasta macOS build is done on macOS 10.15, and that build works on both macOS 10.15 and 10.14. However it does not work on macOS 11 (Big Sur). We need a port of Shasta to macOS 11. This requires:

- Make changes in `shasta/scripts/InstallPrerequisites-macOS.sh`, conditional to macOS version, to make sure all dependencies continue to build and install on both macOS 10.15 and macOS 11.
- Make any code changes (also, if necessary, conditional to macOS version) necessary for the build on macOS 11 to be successful.
- Add to `shasta/.github/workflows/Build.yml` a build step for macOS 11.
- Additional work may be necessary to make sure everything works on both hardware architectures used by macOS 11 (`x86_64` and `arm64`). But it would still be a nice result if the above work is done for only one of those two architectures.

If somebody decides to take over this port, I will help - however I have no knowledge of macOS nor a macOS machine to test on. Please comment here to communicate issues that you bump into or questions you may have.

AustinHartman:
I was able to build locally and run a small assembly on macOS 11. Generating a local build only required switching off the `spoa_optimize_for_native` option. However, I ran into an issue when adding the build step. It seems that macOS 11 virtual environments are not available at the moment (see the note here: [https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on))

paoloczi:
Thank you sou much for looking at this - this is good to know. So it seems that we can support macOS 11 as soon as GitHub Actions allows macOS 11 builds.

I assume your test was on Intel hardware - correct? A separate build will be needed to build for Apple ARM hardware. Presumably GitHub Actions will provide a second environment for that.

paoloczi:
If you create a PR with just the changes needed to build, and without the new GitHub Actions job, I will merge it. 

Or perhaps include in the PR the new GitHub Actions job, but commenting it out. That way it will be easy to activate when the new environment becomes available, and in the meantime people will be able to build it themselves.

AustinHartman:
Actually my local build and test was on ARM hardware. Unfortunately, I don't have access to macOS 11 machine using intel hardware, so I'm unable to test that - hopefully GHA provide an environment for each. For now, I'll open a PR with the changes which allowed me to build locally.

[...]

PR #242, opened 2021-03-18
building for macOS 11

AustinHartman:


scripts/InstallPrerequisites-macOS.sh
```diff
@@ -36,7 +36,7 @@ tar -xvf spoa-v3.4.0.tar.gz
 # Build the static library.
 mkdir build-static
 cd build-static
-cmake ../spoa-v3.4.0 -DBUILD_SHARED_LIBS=OFF
+cmake ../spoa-v3.4.0 -DBUILD_SHARED_LIBS=OFF -Dspoa_optimize_for_native=OFF
```
paoloczi, line 39: This way we lose this optimization for macOS 10, but since macOS 10 on its way out and the Shasta macOS version is provided for convenience rather than performance, I agree that we don't need to worry about this.

paoloczi:
See #229 for some discussion about this.

Thank you!

PR #253, opened 2021-05-18
Export method for marker alignments

rlorigro:
This method writes out all the marker alignments to files in `[assemblyDirectory]/Alignments/` containing the following data:
```
kmerId,ordinal0,ordinal1,rlePosition0,rlePosition1,
```


rlorigro:
accidentally added the Bandage script to this PR, I can edit/redo if you want

scripts/generateBandageLabelsFromAlignment.py
```diff
@@ -0,0 +1,117 @@
...
+        return "hot pink"
+    elif refName == "chr22":
+        return "light salmon"
+    elif refName == "chrX":
+        return "yellow green"
+    elif refName == "chrY":
+        return "olive drab"
+    else:
+        print("Warning using gray for unknown chromosome: " + refName)
+        return "gray"
```
paoloczi, line 56: Instead of returning gray, consider generating a color in HSL space with random hue and fixed saturation/lightness. That would make the script more usable for non-human genomes, or for human genomes where chromosomes are labeled differently.

scripts/generateBandageLabelsFromAlignment.py
```diff
@@ -0,0 +1,117 @@
+from collections import defaultdict
```
paoloczi, line 1: According to [this](https://docs.python.org/3/library/collections.html), the `collections` module is deprecated and will be removed in Python 3.10 (and we are at 3.8). Consider using the replacement instead, `collections.abc`.

rlorigro: I tried to find some more information about this, and I believe the deprecation warning does not actually apply to `defaultdict`:
https://stackoverflow.com/questions/60440157/is-there-a-replacement-for-python-namedtuple

If it did, I believe Python would have given me a warning during run time. 

You can see from [here](https://docs.python.org/3.5/library/collections.abc.html#collections-abstract-base-classes) that there used to be an abc submodule that belonged to `collections`:

> New in version 3.3: Formerly, this module was part of the collections module.

rlorigro: Admittedly very confusing though

scripts/generateBandageLabelsFromAlignment.py
rlorigro: In the end I disallowed all non-conventional chromosome names on line 70 during the counting loop because the decoy/alt chromosomes in the reference were annoyingly attracting a lot of alignments

PR #258, opened 2021-06-18
Fix Data access for browser AlignmentCandidate graph

rlorigro:
Somehow during testing the necessary data objects for building/labeling the local AlignmentCandidate graph were loaded already, so I had not previously encountered the case where they needed to be accessed. This addresses that case.


paoloczi:
This should go in `Assembler::accessAllSoft`, not here.

rlorigro:
Ok I added the CandidateTable access method to that method, since it was the only one missing. The page works now.

PR #283, opened 2022-01-27
Update Commands.html

0seastar0:


docs/Commands.html
```diff
@@ -53,19 +53,19 @@ <h3 id=assemble>Command <code>assemble</code></h3>
...
 used to store data in the filesystem.
 
 <p>
-This also removes the <code>Data</code> directory.
+It also removes the <code>Data</code> directory.
 If a <code>DataOnDisk</code> directory exists,
-it creates a symbolic link to it, named <code>Data</code>.
-This makes it possible to use Shasta http server functionality
-and the Python API even after the binary data are 
+it creates a symbolic link named <code>Data</code>.
```
paoloczi, line 66: This new wording does not say what the symbolic link points to.

docs/Commands.html
```diff
@@ -89,25 +89,23 @@ <h3>Command <code>createBashCompletionScript</code></h3>
 
 <h3>Command <code>explore</code></h3>
 <p>
-This command starts Shasta in a mode in which it behaves as an
+This command starts Shasta in a mode that behaves as an
```
paoloczi, line 92: Similar comment to a similar change yesterday. The new wording does not sound right to me.

docs/Commands.html
```diff
@@ -158,8 +156,8 @@ <h3>Command <code>listConfigurations</code></h3>
 
 <h3 id=saveBinaryData>Command <code>saveBinaryData</code></h3>
 <p>
-This is used to save Shasta binary data.
-Binary data are stored in directory <code>Data</code> in the assembly directory.
+This command is used to save Shasta binary data.
+Shasta stores binary data in directory <code>Data</code> in the assembly directory.
 This command makes a copy on disk in <code>Data</code>.
```
paoloczi, line 161: Actually the copy on disk is named `DataOnDisk`, not `Data`. This error was already there before these changes.

docs/Commands.html
0seastar0: it is correct :)

docs/Commands.html
0seastar0: Ok. I will update it with DataOnDIsk

docs/Commands.html
0seastar0: I will leave the original wording

PR #284, opened 2022-03-05
Update README.md to fix broken link to assemblies

0seastar0:


README.md
```diff
@@ -10,7 +10,7 @@ for error analysis of the Shasta assembler and more.
 Reads from this paper are available 
 [here](https://s3-us-west-2.amazonaws.com/human-pangenomics/index.html).
 The assembly results are
-[here](https://s3-us-west-2.amazonaws.com/human-pangenomics/index.html?prefix=assemblies).
+[here](https://s3-us-west-2.amazonaws.com/human-pangenomics/index.html?prefix=publications/SHASTA2019/assemblies/raw/shasta/).
```
paoloczi, line 13: This is a needed change as a result of some restructuring that must have been done without notifying me, and thank you for making it, but wouldn't it be even better to link to https://s3-us-west-2.amazonaws.com/human-pangenomics/index.html?prefix=publications/SHASTA2019/assemblies/ instead? That way they also get access to the polished and scaffolded assemblies, plus the assemblies done with other assemblers for comparison.

0seastar0: Sounds good.

PR #287, opened 2022-04-07
Update QuickStart.html with latest releases

0seastar0:


docs/QuickStart.html
```diff
@@ -62,13 +62,13 @@ <h2 id="QuickStartMacOS">macOS</h2>
 
 <pre>
 # Download the executable for the latest release.
```
paoloczi, line 64: For macOS, 0.9.0 uses different names for the executables, to distinguish between the Intel version and the ARM version. See [here](https://github.com/chanzuckerberg/shasta/releases/tag/0.9.0) for details.

0seastar0: Do you want to leave it as is or change it and add this text? Not sure how you want to proceed.

paoloczi: If a user were to follow the directions in this PR as currently written, the download would fail because an executable with that name does not exists. The instructions before the PR work, but download 0.8.0.

So I think we have two choices:

1. Leave everything as is, which results in people downloading 0.8.0.

2. Add text to explain that  there are two macOS versions, and you have to pick the one consistent with your processor. Then duplicate the quick start guide, specializing each copy to Intel or ARM processors respectively.

0seastar0: We can resolve this in the next release.

0seastar0: We can resolve this in the next release.

0seastar0:
Since the next release is ~2-3 weeks away, we can update QuickStart.html with the newest info at that time.

paoloczi:
Sounds good, but I want to caution people reading this that the timing of the next release is tentative and subject to change. So please don't make any plans based on the above estimate.

Is it ok to delete the branch?

